PR #13, opened 2019-03-29
Update site in prep for beta

ameyer-pivotal:
- Update homepage
- Add tutorial
- Polish docs (sync with pack readme)

Signed-off-by: Andrew Meyer <ameyer@pivotal.io>


docs/CNAME
```diff
@@ -1 +0,0 @@
-buildpacks.io
```
ameyer-pivotal: Not sure why this happened. Reverting...

ameyer-pivotal: Basically, there are top-level directories called `docs` and `resources`. I'm unsure if they're actually needed or just leftover. I simply copied the real `docs` and `resources` dirs that are located in `src` and replaced the outer ones with those too. I'm not sure this is the right thing to do, so some guidance would be helpful.

This deletion is a result of this file not existing in the original dir I copied from...

sclevine:
@ameyer-pivotal I think I fixed it. Just FYI, to build:
```
$ cd src
$ hugo
```

jkutner:
In 8c93090 i made changes to the header and title. Happy to revisit though.

image

PR #15, opened 2019-04-04
Fix "default start command" example

michaeldwan:


elbandito:
@michaeldwan while you're at it, I think this also needs to be updated:
https://github.com/buildpack/docs/blob/master/docs/docs/create-buildpack/make-app-runnable/index.html#L139

michaeldwan:
@elbandito good catch, fixed

ameyer-pivotal:
@michaeldwan We've made some adjustments that make our repo simpler. Unfortunately, it requires redoing this change.

jkutner:
This was fixed elsewhere

PR #18, opened 2019-04-24
Fix grammar you're -> your

parndt:


ameyer-pivotal:
@parndt We've made some adjustments that make our repo simpler. Unfortunately, it requires redoing this change. We'd do it ourselves, but don't want to steal your thunder. Good catch!

parndt:
@ameyer-pivotal there you go

ameyer-pivotal:
Thanks @parndt!

PR #19, opened 2019-04-26
Add some documentation for using environment variables

quorauk:


ameyer-pivotal:
@quorauk We had actually ported over the env var docs that were in the [`pack`](github.com/buildpack/pack) README fairly recently, so this is no longer necessary.

On top of that, we've simplified/restructured this repo significantly (generated html is no longer under version control), but it unfortunately requires some rework on most PRs. I'll go ahead and close this PR, but if you'd like to suggest rewording or updates to the env var section, by all means feel free via another PR.

Issue #20, opened 2019-04-26
Add some documentation for contributing to this repository


quorauk:
It might be useful to have some docs for contributing to the documentation.

Perhaps something like the following in README.md

## Development
This documentation is build using hugo, install it using the guide [here](https://gohugo.io/getting-started/installing/)
```
git clone https://github.com/buildpack/docs.git
cd src
hugo server
# documentation is available at localhost:1313
```

Zanadar:
Should be covered by https://github.com/buildpacks/docs#development
Feel free to re-open @quorauk if you disagree

PR #21, opened 2019-04-26
Add instructions for editing documentation

quorauk:
addresses #20


ameyer-pivotal:
@quorauk  We've made some adjustments that make our repo simpler. Unfortunately, it requires redoing this change.

Note that part of the refactor actually added some of the basic hugo commands to the README, but I still think the additional info you provide is useful.

PR #23, opened 2019-04-30
fixed double quotes

ktpv:


ameyer-pivotal:
@ktpv We've made some adjustments that make our repo simpler. Unfortunately, it requires redoing this change.

Your original change was in the rendered html (which no longer exists, as it's generated). You'll likely want to make the change in the markdown instead.

ameyer-pivotal:
Also looks like this may be duplicated by https://github.com/buildpack/docs/pull/15

jkutner:
This was fixed elsewhere

PR #25, opened 2019-05-28
fix full path to files

drnic:
Signed-off-by: Dr Nic Williams <drnicwilliams@gmail.com>


drnic:
And this fix is needed for https://buildpacks.io/docs/create-buildpack/detection/

ameyer-pivotal:
@drnic We've made some adjustments that make our repo simpler. Unfortunately, it requires redoing this change.

jkutner:
Fixed in 2fb0864

PR #30, opened 2019-05-28
Backtick `build` on make-app-runnable page

drnic:
Signed-off-by: Dr Nic Williams <drnicwilliams@gmail.com>


ameyer-pivotal:
@drnic We've made some adjustments that make our repo simpler. Unfortunately, it requires redoing this change.

PR #49, opened 2019-09-01
[rebase] explicit note that publish will make change directly to registry nicely

drnic:
from https://buildpacks.slack.com/archives/C94UJCNV6/p1567376786067800

### DIFF
--- a/content/docs/using-pack/update-app-rebase.md
+++ b/content/docs/using-pack/update-app-rebase.md
@@ -26,8 +26,7 @@ run image called `pack/run`. Running the following will update the base of `my-a
 $ pack rebase my-app:my-tag
 ```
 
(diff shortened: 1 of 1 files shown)


sclevine:
If we wanted to explain the implementation of `pack rebase --publish` in more technical detail, we should specify which metadata are being manipulated on the registry (image manifests), have similar documentation for `pack build --publish`, and provide more details about the motivation and performance implications. Otherwise, I think this change will lead to more confusion than the original text.

I do think it's a good idea to explain these things, but maybe we could create a new page for it, and link the original text to the new page?

jromero:
@drnic saw that this PR has a few conflicts, in trying to resolve them I created a new PR that hopefully satisfies this request while generalizing the statement for the time being until we have a more detailed area to put this information. See #126

drnic:
Thanks

PR #57, opened 2019-09-07
Add References docs

### LINKED ISSUE #11: Add reference documention
To make it easier for buildpack authors to create buildpacks, we should add a reference section that describes the interface for each buildpack script (in less formal language, compared to the spec).

jkutner:
Fixes #11

I'm not sure if the Platform and Distribution pages are really needed. But I added them for completeness.

### DIFF
--- a/content/docs/reference/_index.md
+++ b/content/docs/reference/_index.md
@@ -8,7 +8,8 @@ lastmodifieremail = "jpkutner@gmail.com"
 +++
 
 This section provides an overview of the Cloud Native Buildpack API specification. Most
-buildpack users won't need this information unless they are writing a buildpack.
+buildpack users won't need this information unless they are writing a buildpack or
+a platform that supports buildpacks.
 
 The Cloud Native Buildpack API specification consists of the following parts:
 
--- a/content/docs/reference/_index.md
+++ b/content/docs/reference/_index.md
@@ -1,0 +1,17 @@
++++
+title="Reference"
+weight=2
+creatordisplayname = "Joe Kutner"
+creatoremail = "jpkutner@gmail.com"
+lastmodifierdisplayname = "Joe Kutner"
+lastmodifieremail = "jpkutner@gmail.com"
++++
+
+This section provides an overview of the Cloud Native Buildpack API specification. Most
+buildpack users won't need this information unless they are writing a buildpack.
+
+The Cloud Native Buildpack API specification consists of the following parts:
+
+* [Buildpack API](/docs/reference/buildpack-api) - defines the interface between a buildpack and the environment that runs a buildpack.
+* [Platform API](/docs/reference/platform-api) - defines the interface between the environment that runs a buildpack a platform that supports it.
+* [Distribution API](/docs/reference/distribution-api) - defines the artifact format, delivery mechanism, and order resolution process for buildpacks.
--- a/content/docs/reference/_index.md
+++ b/content/docs/reference/_index.md
@@ -1,6 +1,6 @@
 +++
 title="Reference"
-weight=2
+weight=99
 creatordisplayname = "Joe Kutner"
 creatoremail = "jpkutner@gmail.com"
 lastmodifierdisplayname = "Joe Kutner"
--- a/content/docs/reference/buildpack-api.md
+++ b/content/docs/reference/buildpack-api.md
@@ -132,9 +132,14 @@ identifier.
 ### Example
 
 ```
+api = "0.2"
+
 [buildpack]
 id = "example.com/python"
 version = "1.0"
+
+[[stacks]]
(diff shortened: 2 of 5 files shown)


content/docs/reference/_index.md
```diff
@@ -0,0 +1,18 @@
++++
+title="Reference"
+weight=2
```
jromero: Because our docs/site is more centered toward newcomers I would think this would come last in the menu. Similar to how it is on the index (docs) page. Changing the weight to `4` would make it consistent (aka last on the menu).

content/docs/reference/buildpack-api.md
```diff
@@ -0,0 +1,142 @@
...
+
+## `buildpack.toml`
+
+A buildpack must contain a `buildpack.toml` file in its root directory. This
+descriptor file must contain a `buildpack.id` that is a globally unique
+identifier.
+
+### Example
+
+```
```
jromero: We're missing `api = "0.2"` here.

content/docs/reference/buildpack-api.md
```diff
@@ -0,0 +1,142 @@
...
+[buildpack]
+id = "example.com/python"
+version = "1.0"
+```
```
jromero: `Pack` has validation that makes it a requirement to have EITHER `stacks` or `order`. I don't see that restriction explicitly in the spec but instead...

> If an order is specified, then stacks MUST NOT be specified.

https://github.com/buildpack/spec/blob/master/buildpack.md#buildpacktoml-toml

We probably need to align something here.

content/docs/reference/_index.md
jkutner: totally agree. i was wondering how the order was determined! i'll fix that

content/docs/reference/buildpack-api.md
```diff
@@ -0,0 +1,147 @@
...
+### Example
+
+This is a simple example of a buildpack that detects a Python application by
+checking for the presence of a `requirements.txt` file:
+
+```
+#!/bin/sh
+
+if [ -f requirements.txt ]; then
+  echo "Python"
```
ameyer-pivotal: Maybe say "Python Buildpack" here to make clearer that it's the buildpack name you're printing and not the language?

content/docs/reference/buildpack-api.md
```diff
@@ -0,0 +1,147 @@
...
+LAYERS_DIR="$1"
+
+PIP_LAYER="$LAYERS_DIR/pip"
+mkdir -p "$PIP_LAYER/modules" "$PIP_LAYER/env"
+
+pip install -r requirements.txt -t "$PIP_LAYER/modules" \
+  --install-option="--install-scripts=$PIP_LAYER/bin" \
+  --exists-action=w --disable-pip-version-check --no-cache-dir
+
+echo "launch = true" > $PIP_LAYER.toml
```
ameyer-pivotal: Minor: Does this env var need quotes around it, for consistency?

content/docs/reference/platform-api.md
```diff
@@ -0,0 +1,31 @@
++++
```
ameyer-pivotal: The details in this and the distribution doc seem a bit sparse compared to the buildpack one, but it's still helpful to have them.

content/docs/reference/distribution-api.md
```diff
@@ -0,0 +1,25 @@
...
+
+This specification defines the artifact format, delivery mechanism, and order resolution process for buildpacks.
+
+## Buildpack Descriptor
+
+A buildpack must contain a `buildpack.toml` file in its root directory.
+
+## Buildpackage
+
+A buildpackage is a distributable artifact that contains a buildpack. Its format may be either:
```
ameyer-pivotal: Can it contain multiple buildpacks? (Or perhaps if the answer is: yes, via an order buildpack, then maybe we should explain how to do that in the buildpack doc).

ameyer-pivotal:
May want this to go into `edge` rather than `master`, in case it gets merged before we release (or we could just wait until after the release to merge as is).

content/docs/reference/platform-api.md
jkutner: I didn't want to copy/paste the spec, so I tried to narrow this down to essentials (and things that wouldn't change frequently.

jkutner: that said, I'm happy to add more if you think we should

content/docs/reference/distribution-api.md
jkutner: I was hoping to avoid any complex discussion on order buildpacks here by referring to the singular "buildpack" (which is still accurate). But maybe it's just more confusing without a mention?

ameyer-pivotal: That makes sense

PR #70, opened 2019-10-08
Add blurb about api value in buildpack.toml

danielleadams:
Adds more explanation of `api` in `buildpack.toml`, from this issue: https://github.com/buildpack/pack/issues/306

Specifically requested here: https://github.com/buildpack/docs/issues/66

### DIFF
--- a/content/docs/reference/buildpack-api.md
+++ b/content/docs/reference/buildpack-api.md
@@ -129,7 +129,7 @@ A buildpack must contain a `buildpack.toml` file in its root directory. This
 descriptor file must contain a `buildpack.id` that is a globally unique
 identifier.
 
-The `buildpack.toml` has an `api` value that indicates the implemented buildpack API. The format should be `<major>.<minor>` or `<major>`, where `<major>` is equivalent to `<major>.0`. This value is optional and defaults to `0.1`.
+The `buildpack.toml` has an `api` value that indicates the implemented Buildpack API. The format should be `<major>.<minor>` or `<major>`, where `<major>` is equivalent to `<major>.0`. With a pre-stable Buildpack API version, in that `<major>` equals `0`, minors must be equal. If `<major>` is greater than `0`, `<minor>` must be less than or equal to Buildpack API `<minor>`. Major versions must always be equal. The current Buildpack API version is `0.2`.
 
 ### Example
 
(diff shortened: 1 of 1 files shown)


content/docs/reference/buildpack-api.md
```diff
@@ -129,6 +129,8 @@ A buildpack must contain a `buildpack.toml` file in its root directory. This
 descriptor file must contain a `buildpack.id` that is a globally unique
 identifier.
 
+The `buildpack.toml` has an `api` value that indicates the implemented buildpack API. The format should be `<major>.<minor>` or `<major>`, where `<major>` is equivalent to `<major>.0`. This value is optional and defaults to `0.1`.
```
jromero: We probably want to omit optionality. `pack` currently defaults to 0.1 when not preset for backwards compatibility but it's not something expected to be the same across platforms and able to change in the future. It's thereby not specified in the spec.

Other related information worth detailing but not necessary to accept this PR:
- Instead of specifying what the _default version_ pack uses, specify the _current Buildpack API version_ (0.2).
- Explain the following portion of the spec (the [logic](https://github.com/buildpack/pack/blob/master/api/version.go#L80-L82) and [tests](https://github.com/buildpack/pack/blob/master/api/version_test.go#L34) for this if it help)
> When `<major>` is `0`, the buildpack is only compatible with lifecycles implementing that exact buildpack API.
> When `<major>` is greater than `0`, the buildpack is only compatible with lifecycles implementing buildpack API `<major>.<minor>`, where `<major>` of the lifecycle equals `<major>` of the buildpack and `<minor>` of the lifecycle is greater than or equal to `<minor>` of the buildpack.

danielleadams: Thanks for the feedback @jromero. I tried to explain the versioning a little more doc friendly.. if it's too verbose though I can just remove it since the current version is only at 0.2 and may not need the whole explanation yet.

jromero: It looks good IMO. We can certainly iterate over it as well. For now, it's a lot better versus having a void.

danielleadams:
@jromero Thanks 👍 is there a build/deploy step I need to do, or can I just merge this?

jromero:
@danielleadams seems like you've got the permissions necessary to merge. I would say it's safe to merge. 

Deployment happens automatically after merging to master.

PR #85, opened 2019-11-11
Add anchor links for headers

danielleadams:
This adds links to all `h2`, `h3`, `h4` elements for anchor links as suggested in https://github.com/buildpack/docs/issues/65.

I _think_ this works, but locally the anchor links weren't anchoring to the correct locations, so it would be useful to test this on some sort of staging environment if that exists.

### DIFF
--- a/themes/buildpacks/assets/scss/components/_docs.scss
+++ b/themes/buildpacks/assets/scss/components/_docs.scss
@@ -71,6 +71,20 @@
       a {
         text-decoration: none;
         border-bottom: 1px solid rgba($blue-dark, 0.25);
+        &.h-anchor {
+          margin-left: -0.9em;
+          border-bottom: 0px;
+          i {
+            font-size: 0.7em;
+            visibility: hidden;
+            border-bottom: 0px;
+          }
+          &:hover {
+            i {
+              visibility: visible;
+            }
+          }
+        }
       }
     }
 
--- a/themes/buildpacks/layouts/_default/baseof.html
+++ b/themes/buildpacks/layouts/_default/baseof.html
@@ -18,6 +18,27 @@
       gtag('config', 'UA-124323167-1');
     </script>  */}}
 
+    <script>
+      function addAnchor(element) {
+        if (element.id === "" || element.getElementsByTagName("a").length > 0) {
+          return;
+        }
+
+        element.innerHTML =
+          `<a href="#${element.id}" class="h-anchor">
+            <i class="fas fa-link icon" ariaLabel="Anchor"></i> ${element.innerHTML}
+          </a>`;
+      }
+
+      document.addEventListener('DOMContentLoaded', function () {
+        var headers = document.querySelectorAll('h1, h2, h3, h4, h5');
+
+        if (headers) {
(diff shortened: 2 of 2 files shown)


jromero:
Thanks for taking this on!

Unfortunately we don't have a staging server. That's a good suggestion.

---

Looked at it real quick and think I might have found the issue. It looks related to this: https://discourse.gohugo.io/t/hugo-is-generating-css-ids-for-heading-elements-should-it/3284/11

Basically, the auto-generated id for the header element is changed when you add a link creating a now out-of-sync hash link.

without (master): 
image

with (this branch):
image


### Other feedback

There are probably some areas where you want to disable this feature such as: http://localhost:1313/docs/concepts/

image

Maybe a blanket condition of "if header has link that doesn't start with a hash add icon" would suffice.

content/docs/concepts/components/builder.md
```diff
@@ -7,21 +7,21 @@ aliases=[
...
+## [Anatomy of a builder](#anatomy-of-a-builder)
 
 A builder consists of the following components:
 
 * [Buildpacks][buildpack]
 * [Lifecycle][lifecycle]   
 * [Stack's][stack] build image   
 
-### Resources
+### [Resources](resources)
```
jromero: Missing hash `#`

themes/buildpacks/assets/scss/components/_docs.scss
```diff
@@ -70,7 +70,17 @@
       }
       a {
         text-decoration: none;
-        border-bottom: 1px solid rgba($blue-dark, 0.25);
+        margin-left: -1.5rem;
+        .h-anchor {
+          font-size: 1.25rem;
```
jromero: Opinion: The icon looks a little large in certain areas. It should probably scale based on header size. Maybe we want `em` here instead of `rem`.

Ref: https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units#Relative_length_units

danielleadams: sounds good. I used `rem` originally so I could use a set `margin-left` for the header, but I can play around with it

danielleadams:
@jromero I think it would be better to have all h2, h3, and h4 headers have links without having to be added at build time - I just don't know how it would work with some of the headers that have characters that are not alphanumeric (use regex?).

jromero:
@danielleadams The IDs would already be generated so you could simply retrieve them from the header (`hX`) elements. 

Without digging to deep, what I was thinking was to use JS to:

1. gather all `hX` tags that aren't wrapped in a link `a`
2. wrap them in a link pointing to the id of the header:
`<a href='#<id of hX>' class='h-anchor'><hX id="<id of hX>" .../></a>`
3. prepend the anchor icon via css `.h-anchor::before`

Just an idea.

danielleadams:
@jromero made the requested changes. I tried using some of the browser APIs (`::before` and `.insertAdjacentHTML()`, but since it's a font awesome icon, it didn't play nicely)... using `.innerHTML = ` is hacky, but it seems to work nicely for what we want. Let me know what you think.

PR #86, opened 2019-11-12
Added an explicit directory before the pack create-builder command

jkutner:
This threw me off when following through the guide.


jromero:
I think the confusion must be elsewhere because the tutorial asks you to create a `builder.toml` so if you follow the instructions it would use the file you created not necessarily the one in a different directory.

I do think that we could go without step 0 and instead provide hosted archived buildpacks somewhere. Maybe github releases. Then you don't have all the samples to potentially get confused by.

jkutner:
ah my bad. i read it in a hurry and assumed it was using the samples to start

PR #87, opened 2019-11-18
Clear up Buildpack API nomenclature

### LINKED ISSUE #66: Document Buildpack API Version
Request based on buildpack/pack#306

It should probably live here: https://buildpacks.io/docs/reference/buildpack-api/

Reference: https://github.com/buildpack/spec/pull/61/files

### LINKED ISSUE #71: More explanation of sections in buildpack.toml
Based on this issue about adding buildpack API to the docs (https://github.com/buildpack/docs/issues/66), it may be helpful to have more documentation for the `buildpack.toml` for each section (ie. id, stack, metadata, etc).

Looking here at `buildpack.toml` section: https://buildpacks.io/docs/reference/buildpack-api/

### LINKED ISSUE #82: Update Buildpack Author's Guide to API v3
The ["Building blocks of a Cloud Native Buildpack"](https://buildpacks.io/docs/buildpack-author-guide/create-buildpack/building-blocks-cnb/) section of the Buildpack Author's Guide uses the v2 Buildpack API.

Update the guide to use the v3 API.

jromero:
* Rename/Remove Buildpack API v3 mentions
* Define buildpack.toml schema
* Define API compatibility rules

Resolves #82
Resolves #71
Resolves #66

### Before
image

### After
image

<hr />

image

### DIFF
--- a/content/docs/reference/buildpack-api.md
+++ b/content/docs/reference/buildpack-api.md
@@ -144,7 +144,7 @@ The schema is as follows:
 - **`api`** _(string, required, current: `0.2`)_\
     The Buildpack API version the buildpack adheres to. Used to ensure [compatibility](#api-compatibility) against
     the [lifecycle][lifecycle].
-    
+
     > Not to be confused with Cloud Foundry or Heroku buildpack versions. This version pertains to the interface
     > between the [buildpack][buildpack] and the [lifecycle][lifecycle] of Cloud Native Buildpacks.
 
@@ -153,37 +153,41 @@ The schema is as follows:
 
     - **`id`** _(string, required)_\
     A globally unique identifier.
-    
+
     - **`version`** _(string, required)_\
     The version of the buildpack.
     
+    - **`name`** _(string, required)_\
+    Human readable name.
+
     - **`clear-env`** _(boolean, optional, default: `false`)_\
     Clears user-defined environment variables when `true` on executions of `bin/detect` and `bin/build`.
 
 - **`stacks`** _(list, optional)_\
     A list of stacks supported by the buildpack.
     _If omitted, `order` list must be present. Cannot be used in conjunction with `order` list._
-    
+
     - **`id`** _(string, required)_\
     The id of the supported stack.
-    
+
     - **`mixins`** _(string list, required)_\
     A list of mixins required on the stack images.
-    
+
 - **`order`** _(list, optional)_\
-  A list of buildpack groups. This list determines the order in which groups of buildpacks
-  will be tested during detection. For the purpose of creating a [meta-buildpack][meta-buildpack]. _If omitted, `stacks` list must be present. Cannot be used in conjunction with `stacks` list._
+  A list of buildpack groups for the purpose of creating a [meta-buildpack][meta-buildpack]. This list determines the
+  order in which groups of buildpacks will be tested during detection. _If omitted, `stacks` list must be present.
+  Cannot be used in conjunction with `stacks` list._
   
     - **`group`** _(list, required)_\
-    A set of buildpack references.
-    
+    A list of buildpack references.
+
         - **`id`** _(string, required)_\
-          The identifier of a buildpack from the configuration's top-level `buildpacks` list. Buildpacks with the same ID may
-          appear in multiple groups at once but never in the same group.
-        
+          The identifier of a buildpack being referred to.
+          Buildpacks with the same ID may appear in multiple groups at once but never in the same group.
+
         - **`version`** _(string, required)_\
           The version of the buildpack being referred to.
-        
+    
         - **`optional`** _(boolean, optional, default: `false`)_\
           Whether or not this buildpack is optional during detection.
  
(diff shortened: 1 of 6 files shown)


content/docs/reference/buildpack-api.md
```diff
@@ -144,6 +137,89 @@ version = "1.0"
...
+The schema is as follows:
+   
+- **`api`** _(string, required, current: `0.2`)_\
+    The Buildpack API version the buildpack adheres to. Used to ensure [compatibility](#api-compatibility) against
+    the [lifecycle][lifecycle].
+    
+    > Not to be confused with Cloud Foundry or Heroku buildpack versions. This version pertains to the interface
+    > between the [buildpack][buildpack] and the [lifecycle][lifecycle] of Cloud Native Buildpacks.
+
+- **`buildpack`** _(required)_\
```
briandealwis: This `buildpack` section is missing `name` (https://github.com/buildpack/spec/blob/master/buildpack.md#buildpacktoml-toml)

jromero: Good 👀! Added.

content/docs/reference/buildpack-api.md
```diff
@@ -144,6 +137,92 @@ version = "1.0"
...
+
+- **`stacks`** _(list, optional)_\
+    A list of stacks supported by the buildpack.
+    _If omitted, `order` list must be present. Cannot be used in conjunction with `order` list._
+
+    - **`id`** _(string, required)_\
+    The id of the supported stack.
+
+    - **`mixins`** _(string list, required)_\
+    A list of mixins required on the stack images.
```
ameyer-pivotal: Eventually, we can link this to the mixins docs page

content/docs/reference/buildpack-api.md
```diff
@@ -144,6 +137,92 @@ version = "1.0"
...
+
+- **`order`** _(list, optional)_\
+  A list of buildpack groups. This list determines the order in which groups of buildpacks
+  will be tested during detection. For the purpose of creating a [meta-buildpack][meta-buildpack]. _If omitted, `stacks` list must be present. Cannot be used in conjunction with `stacks` list._
+  
+    - **`group`** _(list, required)_\
+    A set of buildpack references.
```
ameyer-pivotal: I would use the word "list" here rather than "set"

content/docs/reference/buildpack-api.md
```diff
@@ -144,6 +137,92 @@ version = "1.0"
...
+
+        - **`id`** _(string, required)_\
+          The identifier of a buildpack from the configuration's top-level `buildpacks` list. Buildpacks with the same ID may
```
ameyer-pivotal: Looks like this was copy/pasted from the builder config page. May want to update the wording in this description. Perhaps:

> The identifier of a buildpack from an external buildpack source (such as a builder or buildpackage). Buildpacks with the same ID may appear in multiple groups at once but never in the same group.

Note: May opt to omit "buildpackage" references until they're well documented/implemented.

content/docs/reference/buildpack-api.md
```diff
@@ -144,6 +137,92 @@ version = "1.0"
...
+ 
+- **`metadata`** _(any, optional)_\
+    Arbitrary data for buildpack.
+ 
+
+## API Compatibility
+
+API compatibility is determined using the logic:
+
+`SUBJECT` is the Buildpack API version of the **buildpack**.\
```
ameyer-pivotal: Could we use `BP` and `LC`, or `BUILDPACK` and `LIFECYCLE` here instead? I found it difficult to continuously translate as I read.

content/docs/reference/buildpack-api.md
jromero: I simplified the description and omitted the mention of _where buildpacks come from_. I think that is different than from the schema and it's more a function of the `meta-buildpack` which is described elsewhere.

content/docs/reference/buildpack-api.md
jromero: I agree that the names used are not optimal but not sure the proposed names would fit any better. I would be concerned that it could be construed as the version of the buildpack or the lifecycle instead of the _buildpack api version_ of the buildpack or lifecycle.

jkutner: Could we borrow some language from https://semver.org/? For example:

> Precedence is determined by the first difference when comparing each of these identifiers from left to right as follows: Major, minor, and patch versions are always compared numerically. Example: 1.0.0 < 2.0.0 < 2.1.0 < 2.1.1. When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version. Example: 1.0.0-alpha < 1.0.0. Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows: identifiers consisting of only digits are compared numerically and identifiers with letters or hyphens are compared lexically in ASCII sort order. Numeric identifiers always have lower precedence than non-numeric identifiers.

jromero: IIRC, semver compatibility is based on a simple comparison based (without considering [constraints](https://jubianchi.github.io/semver-check/#/)) on the precedence you quoted. Our compatibility is slightly more complex to describe.

#### By examples:

In semver as per [MINOR](https://semver.org/#spec-item-7): 1.2.0 IS compatible with 1.3.0. The inverse is also compatible. 1.3.0 IS compatible with 1.2.0.

In our API versions as per [spec](https://github.com/buildpack/spec/blob/master/buildpack.md#buildpacktoml-toml): A buildpack with Buildpack API 1.2 IS compatible with lifecycle with Buildpack API 1.3. The inverse is NOT compatible. A buildpack with Buildpack API 1.3 IS NOT compatible with lifecycle with Buildpack API 1.2.

jromero: Updated. Let me know what you think.

PR #88, opened 2019-11-21
Bundler no longer fails when there are gem updates

danacr:
The buildpack example breaks in the case it finds an existing gemfile, but the checksum doesn't match, so it fails to create the folder.

After adding `|| true`, it no longer fails to build.

Signed-off-by: Dan Acristinii <dan.acristinii@swisscom.com>


jromero:
Makes sense, thanks for the report and taking the time to address it. 

An alternative we used prior in this tutorial is `-p` as in `mkdir -p ...` doesn't fail if directory already exists but we'll take this as a way to fix the problem as it stands.

PR #90, opened 2019-12-03
Document mixins

ameyer-pivotal:
Signed-off-by: Andrew Meyer <ameyer@pivotal.io>
Signed-off-by: Javier Romero <jromero@pivotal.io>

### DIFF
--- a/content/docs/concepts/components/stack.md
+++ b/content/docs/concepts/components/stack.md
@@ -73,13 +73,13 @@ $ pack build example/app
 
 To create a custom stack, simply create customized build and run images containing the following information:
 
-#### Labels
+### Labels
 
-| Name | Description |
-|------|-------------|
-| `io.buildpacks.stack.id` | Identifier for the stack |
+| Name | Description | Format
+|------|-------------|--------
+| `io.buildpacks.stack.id` | Identifier for the stack | String
 
-#### Environment Variables
+### Environment Variables
 
 | Name | Description |
 |------|-------------|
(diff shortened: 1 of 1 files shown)


content/docs/concepts/components/stack.md
```diff
@@ -92,7 +92,57 @@ To create a custom stack, simply create customized build and run images containi
...
+\
+The following rules apply for mixin declarations:
+
+ - `build:`-prefixed mixins may not be declared on a run image
+ - `run:`-prefixed mixins may not be declared on a build image
+ - Unprefixed mixins must be declared on both stack images
+
+#### Example
+
+_Build image:_\
```
jromero: `\` shouldn't be necessary now that it's a code block.

jkutner:
Are mixins implemented?

ameyer-pivotal:
> Are mixins implemented?

@jkutner The documented rules are now enforced on master :)

content/docs/concepts/components/stack.md
```diff
@@ -92,7 +92,57 @@ To create a custom stack, simply create customized build and run images containi
 > `io.buildpacks.stacks.bionic` as its identifier so long as it will work with buildpacks that declare compatibility with the 
 > `io.buildpacks.stacks.bionic` stack.
 
-### Resources
+## Mixins
+
+Mixins provide a way to document OS-level dependencies that a stack provides to buildpacks. Mixins can be provided at build-time
```
briandealwis: This comment is late (sorry), but I find “OS-level dependencies” a bit confusing. “OS-level” can imply a feature of the kernel.  And “dependencies” usually refers to something that is *required* for operation, whereas these mixing document things that are *provided* by the stack.  Might “features” work better?

And it would be helpful to document that
  1. mixins do not cause packages to be installed (this caused me some confusion when I first learned about buildpacks)
  2. there are no explicit semantics associated with to mixin strings, brings the `run:` and `build:` prefixes
  3. any restrictions that there might be in the string. Can they include spaces or colons?

ameyer-pivotal: Thanks for the feedback @briandealwis! We'll definitely incorporate some of these into the docs. We would also take a PR if you're interested in contributing.

PR #92, opened 2019-12-09
Add p flag to mkdir

thephw:
This makes the cache page consistent with the code snippet on the subsequent page of the example.


jromero:
Thank you!

PR #104, opened 2020-01-12
Use headers for top-level keys in builder config documentation

Zanadar:
This allows for linking to specific parts of the docs.
This PR also add clarification about default lifecycle version

### DIFF
--- a/content/docs/reference/builder-config.md
+++ b/content/docs/reference/builder-config.md
@@ -5,91 +5,106 @@ weight=9
 
 A [builder][builder] configuration schema is as follows:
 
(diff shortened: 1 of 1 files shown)


jromero:
The desired outcome didn't come from this. The headers aren't rendered in nested lists and there are some rendered code blocks where there shouldn't be.

Can we revert this until it gets fixed?

Zanadar:
I think if there may be issues with our ‘make serve’ because I was using that to preview and it all looked right

Zanadar:
Filed https://github.com/buildpacks/docs/issues/105

PR #109, opened 2020-01-16
Fix echo in example

Zanadar:


jkutner:
@Zanadar dco plz

PR #116, opened 2020-01-29
Remove refences to pack version

zmackie:
- Until we have tooling that rebuilds the site when new pack releases come out (#115)

Signed-off-by: Zander Mackie <amackie@pivotal.io>

PR #117, opened 2020-02-11
Updates builder docs to use anchor tags

zmackie:
- Blocked by https://github.com/buildpacks/docs/pull/116

### DIFF
--- a/content/docs/reference/builder-config.md
+++ b/content/docs/reference/builder-config.md
@@ -5,91 +5,106 @@ weight=9
 
 A [builder][builder] configuration schema is as follows:
 
-- **`description`** _(string, optional)_
+- #### `description` _(string, optional)_
   <br>
+
   A human-readable description of the builder, to be shown in `inspect-builder` output
   (run `pack inspect-builder -h` for more information).
 
-- **`buildpacks`** _(list, required)_
+- #### `buildpacks` _(list, required)_
   <br>
+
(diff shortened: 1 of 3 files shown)


jromero:
Made a few tweaks to handle the bullet list icon.

image

content/docs/reference/builder-config.md
```diff
@@ -5,92 +5,107 @@ weight=9
...
-     A URL or path to an [archive](#supported-archives). If specified, `version` must not be provided.
+
+    A URL or path to an [archive](#supported-archives). If specified, `version` must not be provided.
 
     > If `version` and `uri` are both omitted, `lifecycle` defaults to the version that was last released
     > at the time of `pack`'s release. In other words, for a particular version of `pack`, this default
-    > will not change despite new lifecycle versions being released.
-  
-#### Supported archives
+    > will not change despite new lifecycle versions being released. The current default lifecycle versin can be [seen here](https://github.com/buildpacks/pack/blob/e4315be24c103e3c5722b08561bf13a55876cbbc/internal/builder/lifecycle.go#L18).
```
jromero: This doesn't look accurate since it's pointing to a very specific commit. Plus pointing to source code is not ideal. 

I'd rather we omit this for now and instead use the same strategy for pack version and inject it into the documentation. _This should be a separate issue._

PR #123, opened 2020-03-09
Document reproducible builds

### LINKED ISSUE #112: Document reproducible builds
### Summary

Create a page to document _reproducible builds_.

Some questions that this document should answer:
- What constitutes as a reproducible build?
- What caveats are there, if any?

### Considerations

- What persona cares most about this?
- Where should it live within buildpacks.io?
- The CNB lifecycle implementation is only one implementation of the spec so this document should call out that it only applies to the CNB lifecycle and is not necessarily spec'd.

### Notes

- Relates to: 
    - Slack conversation: https://buildpacks.slack.com/archives/C94UJCNV6/p1580223671047400
    - https://github.com/buildpacks/lifecycle/pull/226

zmackie:
- Fixes #112

### DIFF
--- a/content/docs/reference/reproducibility.md
+++ b/content/docs/reference/reproducibility.md
@@ -26,7 +26,7 @@ Running `pack build cnbs/sample-hello-world:test --publish` produce a container 
 Inspecting the results of the above command, we see the following output:
 
 ```bash
-$ docker pull cnbs/sample-hello-kotlin:test && docker images --digest
+$ docker pull cnbs/sample-hello-world:test && docker images --digest # Pull remotely created image and view IDs and Digests
 REPOSITORY                                   TAG                 DIGEST                                                                    IMAGE ID            CREATED             SIZE
 sample-hello-world                           test                sha256:9e3cfea3f90fb4fbbe855a2cc9ce505087ae10d6805cfcb44bd67a4b72628641   597c49cae461        40 years ago        95.2MB
 sample-hello-moon-app                        test                <none>                                                                    86aab15e22b8        40 years ago        43MB
@@ -35,7 +35,7 @@ sample-hello-moon-app                        test                <none>         
 ### Consequences and Caveats
 
 There are a couple things to note about the above output:
-- We achieve reproducible builds by "zeroing" various timestamps of the layers that `pack` creates. When images are inspected they may have confusing creation times (40 years ago).
+- We achieve reproducible builds by "zeroing" various timestamps of the layers of the output image. When images are inspected they may have confusing creation times (eg. "40 years ago").
 - The `cnbs/sample-hello-moon:test` image does not have an entry for the "DIGEST" column. This is because the digest is produced from the image's manifest and a manifest is only created when an image is stored in a remote registry.
 
 The CNB lifecycle cannot fix non-reproducible buildpack layer file contents. This means that the underlying buildpack and language ecosystem have to implement reproducible output (for example `go` binaries are reproducible by default). Buildpacks that produce identical layers given the same input could be said to be reproducible buildpacks.
--- a/content/docs/reference/reproducibility.md
+++ b/content/docs/reference/reproducibility.md
@@ -1,0 +1,41 @@
++++
+title="Reproducible Builds"
+weight=9
++++
+
+## Summary
+The Cloud Native Buildpacks project aims to create "Reproducible Builds" of container images. For image creation commands (`create-builder`, `package-buildpackag`, `build`) `pack` aims to create in a reproducible fashion. "Reproducible" is hard to define but we'll do so by example:
+
+---
+Running `pack build` produce a container image with the same image ID (*local* case)
+
+**Given**:
+- A workspace directory containing the same source code
+- A builder image with a *given*  
+- One or more buildpacks that produce identical layers given their input*
+
+---
+Running `pack build --publish` produce a container image with the same image digest (*remote* case)
+
+**Given**:
+- A workspace directory containing the same source code
+- A builder image with a *given*  
+- One or more buildpacks that produce identical layers given their input
+
+### Consequences and Caveats
+
+We achieve reproducible builds by "zeroing" various timestamps of the layers that `pack` creates. When images are inspected (via something like `docker inspect`) they may have confusing creation times:
+
+```bash
+REPOSITORY                                   TAG                 IMAGE ID            CREATED             SIZE
+cnbs/sample-builder                          <none>              def52b23918d        40 years ago        234MB
+sample-kotlin-app                            alpine              45dc2d2681a1        40 years ago        18.9MB
+```
+
+All that said, the CNB lifecycle cannot fix non-reproducible buildpack layer file contents. This means that the underlying buildpack and language ecosystem have to implement reproducible output (for example `go` binaries are reproducible by default).
+
+A local and remote build will not produce the same image digest because:
+- The remote image will have an image digest reference in the `runImage.reference` field in the `io.buildpacks.lifecycle.metadata` label
+- The local image will have an image ID in the `runImage.reference` field in the `io.buildpacks.lifecycle.metadata` label
+
+This occurs because, in the daemon case, the run-image may not have a repository digest reference (if it was created locally). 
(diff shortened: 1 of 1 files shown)


content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,41 @@
++++
+title="Reproducible Builds"
+weight=9
++++
+
+## Summary
+The Cloud Native Buildpacks project aims to create "Reproducible Builds" of container images. For image creation commands (`create-builder`, `package-buildpackag`, `build`) `pack` aims to create in a reproducible fashion. "Reproducible" is hard to define but we'll do so by example:
```
natalieparellano: typo: package-buildpackag

natalieparellano: "aims to create in" - aims to create images?

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,41 @@
...
+
+---
+Running `pack build` produce a container image with the same image ID (*local* case)
```
natalieparellano: produce > produces ?

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,41 @@
...
+
+**Given**:
+- A workspace directory containing the same source code
+- A builder image with a *given*
```
natalieparellano: What is given?

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,41 @@
...
+## Summary
+The Cloud Native Buildpacks project aims to create "Reproducible Builds" of container images. For image creation commands (`create-builder`, `package-buildpackag`, `build`) `pack` aims to create in a reproducible fashion. "Reproducible" is hard to define but we'll do so by example:
+
+---
+Running `pack build` produce a container image with the same image ID (*local* case)
+
+**Given**:
+- A workspace directory containing the same source code
+- A builder image with a *given*  
+- One or more buildpacks that produce identical layers given their input*
```
natalieparellano: I am not quite sure what this means. Can you elaborate?

content/docs/reference/reproducibility.md
natalieparellano: Do you mean to say running pack **multiple times** with the same inputs should produce the same output? I think the "multiple" part might be worth emphasizing in the title.

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,41 @@
...
+
+---
+Running `pack build --publish` produce a container image with the same image digest (*remote* case)
```
natalieparellano: produce > produces

natalieparellano: You say "image digest" here and "image ID" above - are these the same? It also might be worth clarifying how a user can find these values (e.g., `docker inspect`).

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,41 @@
...
+
+**Given**:
+- A workspace directory containing the same source code
+- A builder image with a *given*
```
natalieparellano: What is given?

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,41 @@
...
+Running `pack build --publish` produce a container image with the same image digest (*remote* case)
+
+**Given**:
+- A workspace directory containing the same source code
+- A builder image with a *given*  
+- One or more buildpacks that produce identical layers given their input
+
+### Consequences and Caveats
+
+We achieve reproducible builds by "zeroing" various timestamps of the layers that `pack` creates. When images are inspected (via something like `docker inspect`) they may have confusing creation times:
```
natalieparellano: "layers that `pack` creates" -- is all of the zeroing done by pack? I thought the exporter was responsible for some of this. I think it's worth being precise about this.

content/docs/reference/reproducibility.md
natalieparellano: I see now that these are indeed intended to be different... still I think some additional info on how to view / verify these values would be helpful.

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,41 @@
...
+
+```bash
+REPOSITORY                                   TAG                 IMAGE ID            CREATED             SIZE
+cnbs/sample-builder                          <none>              def52b23918d        40 years ago        234MB
+sample-kotlin-app                            alpine              45dc2d2681a1        40 years ago        18.9MB
+```
+
+All that said, the CNB lifecycle cannot fix non-reproducible buildpack layer file contents. This means that the underlying buildpack and language ecosystem have to implement reproducible output (for example `go` binaries are reproducible by default).
+
+A local and remote build will not produce the same image digest because:
```
jromero: This is very interesting. If true, I think we should dig into it deeper. Why does the local image need to be an image ID instead of a reference? References can existing in the daemon. See `docker tag` command.

zmackie: I think the key here is that local daemon stored images won't have a digest until pushed (this is because of docker):
```bash
○ → docker inspect cnbs/sample-builder:alpine | jq --raw-output '.[0].RepoDigests[]'
cnbs/sample-builder@sha256:9e3cfea3f90fb4fbbe855a2cc9ce505087ae10d6805cfcb44bd67a4b72628641

  |2.6.3| NY-Floater-15565 in ~/workspace
○ → docker inspect cnb-test | jq --raw-output '.[0].RepoDigests[]'
Error: No such object: cnb-test
jq: error (at <stdin>:1): Cannot iterate over null (null)
```

jromero: I did confirm that the `runImage.reference` does change between whether you run `pack build` or `pack build ... --publish`. If we could calculate the image digest of a local image I'm curious why we don't. This is now totally irrelevant to this issue. Just something that was brought to light so thank you.

zmackie:
I created https://github.com/buildpacks/pack/issues/534 as a possible solution to the differences in config b/w locally and remotely created images.

content/docs/reference/reproducibility.md
zmackie: I agree that being precise here is useful, but the fact that pack creates these layers via the lifecycle is getting a bit into the weeds for this is as a first pass.  I was aiming for a higher level overview of the behavior that users will observe. That said, I the goal in getting this out is to point folks who have questions about reproducibility here and then further improve the documentation. I could very well image having a more technical/architectural explanation of this concept on this page at some later point, but I don't know if we need that _right_ now. 

All that said, I'm open to being wrong on that!

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,46 @@
...
+---
+Running `pack build sample-hello-moon:test` multiple times produces a container image with the same image ID (*local* case)
+
+**Given**:
+- A workspace directory containing the same source code
+- The same builder image
+- The same set of buildpacks (see caveat below).
+
+---
+Running `pack build cnbs/sample-hello-world:test --publish` produce a container image with the same image digest (*remote* case)
```
yaelharel: Running `...` **multiple times** produce**s**

content/docs/reference/reproducibility.md
```diff
@@ -0,0 +1,46 @@
...
+The Cloud Native Buildpacks project aims to create "Reproducible Builds" of container images. For image creation commands (`create-builder`, `package-buildpack`, `build`) `pack` creates container images in a reproducible fashion. "Reproducible" is hard to define but we'll do so by example and with a few caveats:
+
+## Examples
+---
+Running `pack build sample-hello-moon:test` multiple times produces a container image with the same image ID (*local* case)
+
+**Given**:
+- A workspace directory containing the same source code
+- The same builder image
+- The same set of buildpacks (see caveat below).
```
yaelharel: Does the workspace directory only contain the source code or also the builder image and the buildpacks?
If the later is true, maybe it's better to write:
```
A workspace directory containing:
- The same source code
- The same builder image
- The same set of buildpacks
```

content/docs/reference/reproducibility.md
natalieparellano: Sounds good, we can always add more detail later :)

content/docs/reference/reproducibility.md
zmackie: I'm using workspace sort of vaguely here. This should just really be `- The same source code`

content/docs/reference/reproducibility.md
zmackie: Good call.

PR #128, opened 2020-04-17
Buildpacks API: BUILD_PLAN is a path

ekcasey:
The existing wording makes it seem like a string containing the contents of the build plan.


quidryan:
Thanks! I ran across this and it was confusing.

PR #133, opened 2020-04-24
Add Build Plan information to reference and BP developer guide

### LINKED ISSUE #67: Document build plan
We need to better explain and document the format of build plan. Particularly, the `provides` and `requires` concepts.

References: https://github.com/buildpack/spec/blob/master/buildpack.md#build-plan-toml

This should entail:
* Explaining it as part of the [tutorial](https://buildpacks.io/docs/create-buildpack/make-buildpack-configurable/)
* Adding it to the [buildpack API reference](https://buildpacks.io/docs/reference/buildpack-api/)

dfreilich:
I'm not sure if I correctly explained the Build Plan, vs the Buildpack Plan. 

Closes #67


content/docs/reference/buildpack-api.md
```diff
@@ -187,13 +186,68 @@ The schema is as follows:
...
+
         - **`optional`** _(boolean, optional, default: `false`)_\
           Whether or not this buildpack is optional during detection.
- 
+
 - **`metadata`** _(any, optional)_\
     Arbitrary data for buildpack.
- 
+
+## Build Plan
```
natalieparellano, line 196: I have to confess, I read through this section and the linked spec documents a few times and I am still confused :/

Would it be possible to expand the example a bit - e.g., I'm assuming that there are two buildpacks in play here (node 10.x and node 12.x) - perhaps we could walk through the `detect` phase as each buildpack is run and show how the build plan is changing and then what the lifecycle is going to do with it? Or maybe talk about the rationale for why each buildpack specifies the `provides` and `requires` that it does? I think I'm still unsure why a buildpack would both provide and require `node`, for example...

dfreilich: I agree. I provided some more specific examples, but I think there still is additional clarifications I could provide. Let me know whether it helped!

content/docs/reference/buildpack-api.md
```diff
@@ -187,13 +186,120 @@ The schema is as follows:
...
+name = "node"
+
+[[requires]]
+name = "node"
+version = "10.x"
+
+[requires.metadata]
+version-source = ".nvmrc"
+
+[[requires]]
```
jromero, line 251: I expected this to need an `[or]` otherwise it would make a statement that it requires a node version or `10.x` AND `12.x`. Is that not the case? (Not saying it's inaccurate. Honestly, I don't know the right answer.)

content/docs/reference/buildpack-api.md
```diff
@@ -187,13 +186,120 @@ The schema is as follows:
...
+### Example
+Let's walk through some possible cases a `node-engine` buildpack may consider:
+
+1.  Nothing in the app explicitly calls out that it is needed
+2.  It is explicitly referred to in some configuration file
+3.  Two different configuration files request different versions of `node`
+
+We will also consider what a `NPM` buildpack may do.
+
+#### 1. No Explicit Request
```
jromero, line 214: > If a required buildpack provides a dependency that is not required by the same buildpack or a subsequent buildpack, the trial MUST fail to detect.

Based on the spec, this would fail detection. Maybe we should side-note it.

content/docs/reference/buildpack-api.md
dfreilich: That's what I would have thought, but not necessarily. The [Paketo Node-Engine CNB](https://github.com/paketo-buildpacks/node-engine/blob/master/node/detect.go) does it this way. I find that clearer than with an `[or]`

content/docs/reference/buildpack-api.md
dfreilich: Makes sense!

dfreilich: @jromero I added a note both by the `node` and `npm` cnb clarifications that running either alone wouldn't work, per the lifecycle

content/docs/reference/buildpack-api.md
dfreilich: @sclevine Would there be a preference between using two different requires, versus using an `[or]`?

jromero: > That's what I would have thought, but not necessarily. The [Paketo Node-Engine CNB](https://github.com/paketo-buildpacks/node-engine/blob/master/node/detect.go) does it this way. I find that clearer than with an `[or]`

That's very interesting. Given that we made mention to `[or]` somewhere above, I'd be interested in seeing a use case example where `[or]` _MUST_ be used for full clarity.

content/docs/reference/buildpack-api.md
```diff
@@ -213,20 +230,55 @@ version = "10.x"
 
 [requires.metadata]
 version-source = ".nvmrc"
+```
 
-[[or]]
+As always, the buildpack `provides` `node`. In this particular case, a version of `node` (`10.x`) is being `requested` in a configuration file (`.nvmrc`), and the buildpack chooses to add an additional piece of metadata (`version-source`), so that it can understand where that request came from.
```
natalieparellano: It's a little odd to me to have `requested` in literal quotes since this doesn't appear to be something that's actually written out anywhere. But I get why you are emphasizing it.

content/docs/reference/buildpack-api.md
```diff
@@ -213,20 +230,55 @@ version = "10.x"
...
+[[requires]]
 name = "node"
 version = "12.x"
 
-[or.requires.metadata]
+[requires.metadata]
 version-source = "buildpack.yml"
 ```
 
+As always, the buildpack provides `node`. In this case, it requests two different things: `node v10.x` and `node v12.x`. During the `build` phase, it will analyze the two `requires`, and decide which (or both) to follow. Alternatively, it could have resolved the appropriate version in the `detect` phase, and only written the corresponding entry to the build plan.
```
natalieparellano, line 260: > decide which (or both) to follow

Can a node-engine buildpack provide two different versions of node? It seems so from what's presented here but maybe I am misunderstanding? This might be good to clarify.

> it could have resolved the appropriate version in the `detect` phase

I would love to know how this is done! This info might be helpful to complete the picture.

content/docs/reference/buildpack-api.md
```diff
@@ -203,6 +203,23 @@ A buildpack can `require` or `provide` multiple dependencies, and even multiple
 The lifecycle uses the `Build Plan` as one element in deciding whether or not a particular list of buildpacks is appropriate, by seeing whether all dependencies required can be provided by that list. See the [spec](https://github.com/buildpacks/spec/blob/master/buildpack.md#phase-1-detection) for particulars on how ordering buildpacks can adjust detection results.
 
 ### Example
+Let's walk through some possible cases a `node-engine` buildpack may consider:
+
+1.  Nothing in the app explicitly calls out that it is needed
+2.  It is explicitly referred to in some configuration file
+3.  Two different configuration files request different versions of `node`
+
+We will also consider what a `NPM` buildpack may do.
```
natalieparellano, line 212: It might be helpful (speaking as someone who doesn't know node) to call out what is the responsibility of the node-engine buildpack vs. the NPM buildpack.

content/docs/reference/buildpack-api.md
```diff
@@ -213,20 +230,55 @@ version = "10.x"
...
+```
+[[requires]]
+name = "node"
+version = "14.1"
+
+[requires.metadata]
+version-source = "package.json"
+```
+
+That `requires` will be used by the lifecycle to decide that this (`node-engine` and `npm`) is the correct set of buildpacks, given that the `node-engine` buildpack `provides` `node`, while the `npm` buildpack `requires` `node`, and therefore all `requires` and `provides` are fulfilled.
```
natalieparellano: Does it not also look at the version being requested? It seems like that information would also be used, no?

content/docs/reference/buildpack-api.md
```diff
@@ -194,7 +194,7 @@ The schema is as follows:
     Arbitrary data for buildpack.
 
 ## Build Plan
-The [Build Plan](https://github.com/buildpacks/spec/blob/master/buildpack.md#buildpack-plan-toml) is a document the buildpacks can use to pass information between the [detect](#bindetect) and [build](#bindetect) phases. The build plan is passed (by the lifecycle) as a parameter to the `detect` and `build` binaries of the buildpack.
+The [Build Plan](https://github.com/buildpacks/spec/blob/master/buildpack.md#build-plan-toml) is a document the buildpacks can use to pass information between the [detect](#bindetect) and [build](#bindetect) phases. The build plan is passed (by the lifecycle) as a parameter to the `detect` and `build` binaries of the buildpack.
 * During the `detect` phase, the buildpack(s) may write something it `requires` or `provides` (or both) into the Build Plan.
```
natalieparellano, line 198: After reading on (assuming I'm understanding correctly) it seems that `requires` is conceptually something that the app OR the buildpack can do, is this right? It might be worth calling out, that was the "ah ha" moment for me!

content/docs/reference/buildpack-api.md
```diff
@@ -203,6 +203,23 @@ A buildpack can `require` or `provide` multiple dependencies, and even multiple
...
+
+#### 1. No Explicit Request
+A `node-engine` buildpack is always happy to `provide` the `node` dependency. The build plan it will write may look something like:
+```
+[[provides]]
+name = "node"
+```
+
+#### 2. One Version Requested
+The buildpack sees in one configuration file (e.g. `.nvmrc`) that it is explicitly requested by the application. Seeing that, it may write the below text to the build plan:
```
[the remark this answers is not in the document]
dfreilich: Thanks for the clarification!

content/docs/reference/buildpack-api.md
dfreilich: Hmm. The buildpack is the only thing that writes to the Build Plan, so by that metric, it is the only thing that `requires`. However, what I tried to bring out through the examples is that the buildpack looks at the application to figure out what exactly it needs. On the simplest level, that means it looks to see if it is a `node`, `Java`, or `Go` application, and ensures that the relevant tool is there to run that application. 

Applications (or more likely, operators) will sometimes use configuration files to ensure that the buildpack will know exactly what version of something to use, or to explicitly request something that the buildpack may not be able to detect. That is kind of how the paketo-buildpacks are configured for node - node files are just plain javascript files, so it is only the presence of a `package.json` or a `yarn.lock` file which says that this is a node application. I think that the paketo-buildpacks's node buildpack doesn't look for `package.json` or `yarn.lock` files, however, because those are external to node (they come from the popular package managers, but aren't really part of the node language specification), and therefore, it would just `provide` by default, without necessary being able to know if it is `required`. 

This example would probably be simpler with a language like `go`, though, where all we'd have to do is look for the presence of a `.go` file, and if we find, the `go cnb` would install `go`. That wouldn't show just what you can do to customize the build plan as much, though... 

Let me know if that made sense!!

content/docs/reference/buildpack-api.md
dfreilich: I agree, and based on [this](https://buildpacks.slack.com/archives/CCP60GJLS/p1588625326288900?thread_ts=1588614869.288600&cid=CCP60GJLS) conversation, I'm not sure if the node-engine cnb should be doing that as well. I'm going to use the `jvm` cnb to illustrate using an `[or]` instead, as they do here: https://github.com/paketo-buildpacks/libjvm/blob/master/detect.go#L28-L44

content/docs/reference/buildpack-api.md
natalieparellano: Ah... this makes sense. I think I was assuming that when a buildpack requires something, it means that that something is required for the buildpack to run, vs. required for the app to run. I think in the end requires is (almost?) always about what the app needs to run. This is the wrong assumption I was making... I think you've done a good job of illustrating this :)

dfreilich:
@ekcasey Does the `jvm` part of this make sense?

content/docs/reference/buildpack-api.md
```diff
@@ -187,13 +186,127 @@ The schema is as follows:
...
+[[or]]
+[[or.provides]]
+name = "jre"
+
+```
+
+The buildpack gives three options to the lifecycle:
+* It can provide a standalone `jre`
+* It can provide a standalone `jdk`
+* It can provide both the `jdk` and `jre`
```
jromero, line 285: If I'm understanding how the TOML translates to to 3 possible options some inline comments would help:

```toml
# option 1 (`jre` and `jdk`)
[[provides]]
name = "jre"

[[provides]]
name = "jdk"

# option 1 (or just `jdk`)
[[or]]
[[or.provides]]
name = "jdk"

# option 2 (or just `jre`)
[[or]]
[[or.provides]]
name = "jre"
```

dfreilich: Yup! I'll add inline comments for that

dfreilich: Added the comments in. Thanks for the clarifiation!

PR #156, opened 2020-06-02
Add packaged buildpack to create-builder config

### LINKED ISSUE #145: Document the use of packages in create-builder
We are able to use packages in our `builder.toml` when creating a builder (with the completion of https://github.com/buildpacks/pack/pull/611),  but our docs seem pretty sparse. We should expand the create builder section to explain how else you can create a builder. 

This should include adding it to the [builder.toml config reference](https://buildpacks.io/docs/reference/builder-config/#supported-archives)

- What are my options when creating a builder?

- [ ] App Developer
- [x] Buildpack Author
- [x] Operator

dfreilich:
Closes #145 

Signed-off-by: David Freilich <dfreilich@vmware.com>

### DIFF
--- a/content/docs/reference/builder-config.md
+++ b/content/docs/reference/builder-config.md
@@ -20,14 +20,12 @@ A [builder][builder] configuration schema is as follows:
 
   - The location of the buildpack. Must be **one** of the following:
 
-      - **`uri`** _(string)_
-        <br>
-        A URL or path to an [archive](#supported-archives), or a path to a directory. If path is relative, it must be relative to the `builder.toml`.
+      - **`uri`** _(string)_\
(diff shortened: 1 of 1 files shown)


dfreilich:
Part of the issues involves supplementing the initial create-builder documentation, but right now we point to the samples repo, which seems like a good pattern. Samples already demonstrates the use of a packaged buildpack image; should I add in another example for a packaged buildpack file, or is there no need, given that we clarify it here?

jromero:
> should I add in another example for a packaged buildpack file, or is there no need, given that we clarify it here?

I would be okay with deferring until someone makes a statement that it isn't clear enough. I wouldn't say it's a must. The hope of moving to the buildpacks registry would also make it less likely that `.cnb` files would be widely distributed.

PR #157, opened 2020-06-03
Change brew tap to buildpacks

dfreilich:
Signed-off-by: David Freilich <dfreilich@vmware.com>

### DIFF
--- a/content/docs/install-pack.md
+++ b/content/docs/install-pack.md
@@ -19,8 +19,7 @@ You can install the most recent version of `pack` (version **{{< latest >}}**) a
 To install `pack` on macOS, the easiest way is to use Homebrew:
 
 ```bash
-brew tap buildpacks/tap
-brew install pack
+brew install buildpacks/tap/pack
 ```
 
 > **Optional:** Enable [auto-completion](#auto-completion)
--- a/content/docs/install-pack.md
+++ b/content/docs/install-pack.md
@@ -19,7 +19,7 @@ You can install the most recent version of `pack` (version **{{< latest >}}**) a
 To install `pack` on macOS, the easiest way is to use Homebrew:
 
 ```bash
-brew tap buildpack/tap
+brew tap buildpacks/tap
 brew install pack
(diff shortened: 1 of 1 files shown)


content/docs/install-pack.md
```diff
@@ -19,7 +19,7 @@ You can install the most recent version of `pack` (version **{{< latest >}}**) a
 To install `pack` on macOS, the easiest way is to use Homebrew:
 
 ```bash
-brew tap buildpack/tap
+brew tap buildpacks/tap
 brew install pack
```
jromero, line 23:
```suggestion
brew install buildpacks/tap/pack
```

Minor: We started doing this on release notes as a way to condense two commands into one. Probably worth doing here  as well.

dfreilich:
@jromero How do we remove the continuous-integration check? Given that we're using github-actions now, it seems weird, and doesn't seem to be running

jromero:
> How do we remove the continuous-integration check? Given that we're using github-actions now, it seems weird, and doesn't seem to be running

It's an admin setting. I went ahead and removed it as soon as I saw it. I will open another issue though that we should probably be testing the build. :D

FYI, it's not running a build via GHA because there is no trigger for `pull_requests`. We'd also want to gate the deploy step behind it being a `push` event to the `master` branch.

jromero:
@dfreilich Looks like DCO failed. When accepting suggestions, you have to DCO sign in the commit message. Sorry for not adding the "Pro-tip". Only way to recover now is ammending + force push.

dfreilich:
@jromero Interesting, I actually tried to accept it with the DCO (as you can see from the commit here https://github.com/buildpacks/docs/pull/157/commits/e7d24f5eed1cdd6ff997c6748099ade6e9b17a85), but it looks like since you suggested and it didn't have your DCO attached, it didn't work?

dfreilich:
@jromero Opened a bug https://github.com/probot/dco/issues/128 about it, but I fixed it manually

PR #161, opened 2020-06-19
Feature/119 add pack docs

### LINKED ISSUE #119: Document pack cli commands
In order to make documentation of the behavior of pack more discoverable (and so that we can more easily point to it), it would be nice if pack's commands were documented. This *could* start as a folder in pack proper that we create during releases and then move here somehow.

Cobra supports this ([POC](https://github.com/buildpacks/pack/pull/399)).

dfreilich:
This PR adds Pack CLI documentation as a reference folder, cataloguing every pack cli command. There is currently no release with the required command (`cmd.NewPackCommand`), so it is currently inoperable (until https://github.com/buildpacks/pack/pull/701 is merged, in some fashion). 

I would combine it with a Github Action to regenerate the CLI commands on a new release - but I assume the action shouldn't commit, but should rather make it a PR...   

Resolves #119


dfreilich:
With the release of pack `v0.12.0`, this is now operable. 

Some images:
Initial (pack CLI is in reference, and not shown):
image

When reference is expanded, Pack CLI is last:
image

The initial Pack CLI page:
image

content/docs/reference/pack/_index.md
```diff
@@ -0,0 +1,5 @@
++++
+title="Pack CLI"
++++
+
+This section provides a description of the Pack CLI commands.
```
jkutner, line 5:
```suggestion
This section describes the Pack CLI commands.
```

content/docs/reference/pack/pack.md
```diff
@@ -0,0 +1,41 @@
++++
+title="pack"
++++
+## pack
+
+CLI for building apps using Cloud Native Buildpacks
+
+### Synopsis
+
+CLI for building apps using Cloud Native Buildpacks
```
jkutner, line 10: Why do all the Synopsis sections repeat the description?

jromero:
I don't think we want the generated pages to be committed at all. They should be generated as part of the build process but excluded from git via gitignore otherwise we're going to have a lot of commits where only generated content is updated.

dfreilich:
But if they are not committed, how will they be part of the docs?

content/docs/reference/pack/pack.md
dfreilich: It looks like Synopsis prints out the `cmd.Long` description, and that just combines the `cmd.Short` with the usage if there is no separate long description.

dfreilich:
@jromero With this being approved, any further feelings on that? Should I not include the generated files, and then only generate them as part of the deployment procedure, or do we want the option to make adjustments to the documentation?

jromero:
@dfreilich 

Still trying to understand how this would work if we don’t have them generated as part of the `make build` process. These are my thoughts:

1. We would want them generated every time there is a new version of pack. This would void any hope of trying to get the docs updated outside of the `pack` release process otherwise they would get stomped on when the next release comes out.
2. Instead of having to create a PR/commit to the docs for pack related docs I think it would be easier to trigger a deployment of the new docs similar to what we do for pack version  which we use in other parts of the docs.

dfreilich:
@jromero That's fair. So add the pack.md docs generated to the .gitignore, add `update-docs` to `make build` (so we don't need to add anything new to the update process, it will anyways update the docs with a new release of pack), and then if we any issues with that we can always change it to have the docs included?

jromero:
Yes. That's my vote.

PR #164, opened 2020-06-24
Add high-level project highlights to homepage

jromero:
# Summary

The goal of this PR is to make it clearer what some of the high-level project features are. For the most part this takes what was previously written in text and adds visual representations of them.

## Before
buildpacks io_

## After
localhost_1313_


yaelharel:
This is a great change!
I think it's worth adding a section on what buildpacks are at the top of this page (before "why cloud native buildpacks").
It looks for me more important than the section "about the project".

jromero:
> It's definitely a significant improvement. I don't love the emphasis on control, especially since that seems pretty close to the second. Maybe we should highlight ease of use more, for both app devs and operators? 

@dfreilich Let's keep chatting. This change is primarily about adding visuals to the content that was already there.

I know the @buildpacks/core-team also has some ideas about re-invisioning the highlights so that it's consistent across the project (eg. conference talks).

PR #178, opened 2020-07-30
Add "features" page

jromero:
## Summary

This PR adds a new page "Features" which includes the following:

1. A grid of features along with a simple description.
2. A comparison matrix between CNB and other alternative tools

This page is linked to from the main menu and below the "Why Cloud Native Buildpacks?" section of the homepage.

### Notes 

- There is more information I want to add here like explanations about some tools that aren't really alternatives such as moby, buildah, kaniko, etc. but I thought I'd piecemeal it to have easier focused discussions if we needed to.
- The features listed here are mostly features that I've gathered through presentations and conversations. I'd like to add a bit more polish and organization in the future.

#### Homepage link

image

#### Page

localhost_1313_features_


content/features.md
```diff
@@ -0,0 +1,50 @@
...
+{{< feature title="Fast" >}}
+Only re-builds and uploads layers when necessary.
+{{</>}}
+
+{{< feature title="Multi-language" >}}
+Supports more than one programming language family.
+{{</>}}
+
+{{< feature title="Multi-process" >}}
+Image can have multiple entrypoints. Think operational modes.
```
jkutner, line 31: This wording is a bit odd. Maybe "Image can have multiple entrypoints for each operational mode."

jkutner:
@jromero I love it! but I think we should try to fit "Reuse" or "Reusability" into the features. Something like "Leverage production-ready buildpacks maintained by the community." The reason I think it's a reminder that, unlike Dockerfile, you likely won't need to write one of these yourself.

I like the comparison matrix!

content/features.md
```diff
@@ -0,0 +1,50 @@
...
+{{< feature title="Build Cache" align="right" >}}
+Robust build caching is used to improve performance.
+{{</>}}
+
+{{< feature title="Composable" >}}
+Multiple buildpacks can be used to create an app image.
+{{</>}}
+
+{{< feature title="Fast" >}}
+Only re-builds and uploads layers when necessary.
```
jkutner, line 23: I wonder if people would read this and think "so what, docker does this with layers too". I'm not sure if there's an easy way to say "Only re-builds layers when necessary, not when something unrelated changes."

jromero: Would this block be one that we can or should replace with "Reusability"?

jkutner: I think "Reusability" is more about reusing the buildpacks, not the layers. i could be wrong though

jromero: Yeah, agreed. I was thinking about replacing this block with "Reusability" not so much because they were related but because we could probably improve the way we want to convey _speed_/_optimizations_ and in the meantime we have a need to fit the Reuse block in.

jromero: Made this change on the latest commit.

PR #184, opened 2020-08-16
Update installation document

inductor:
brew command is available both on macOS and Linux. 

I rewrote the current Linux installation as alternative as using package manager is better than installing the binary into your path directly...


inductor:
I signed it

PR #201, opened 2020-08-26
Add Windows build guide

ameyer-pivotal:
- Adds a Windows-centric guide to the App Developer Guides
- References new guide from the main App Journey tutorial, and from the Linux build guide

Signed-off-by: Andrew Meyer <meyeran@vmware.com>

### DIFF
--- a/content/docs/app-developer-guide/build-a-windows-app.md
+++ b/content/docs/app-developer-guide/build-a-windows-app.md
@@ -1,0 +1,87 @@
++++
+title="Build a Windows app"
+weight=2
+summary="The basics of taking your Windows app from source code to runnable image."
+aliases=[
+    "/docs/using-pack/building-windows-app/"
+]
++++
+
+> **EXPERIMENTAL:**
+>
+> - Please note that **Windows container support is currently experimental**. You may be asked to enable experimental features  in `pack` when running the following commands. Simply follow the on-screen instructions to do so.
+> - If you encounter any problems while experimenting, we'd love for you to let us know by filing an issue on the [pack][pack-issues] or [lifecycle][lifecycle-issues] repo.
+
+### Recommended reading
+
+Before trying out builds for Windows images, we recommend following the [Linux image tutorial][app-journey] under [Getting Started][getting-started]. Don't worry, you can still run it on a Windows OS if you need to -- just make sure to [enable Linux containers][container-mode] for Docker first.
+
+When you're done, head back here.
+
+### Understanding Docker hosts
+
+The `pack` CLI supports using the [`DOCKER_HOST`][docker-env-vars] environment variable. Essentially, this points your local Docker client (e.g. `docker` or `pack` CLI) to a remote Docker daemon. For instance, if the IP address of your host is `192.168.2.100`, you can set `DOCKER_HOST` to `tcp://192.168.2.100`. Any subsequent `pack` or `docker` commands would communicate with the daemon on that machine instead of a local daemon.
+
+By setting `DOCKER_HOST`, you can use `pack` on any OS to build Windows container images, as long as your remote Docker host is configured to support them (i.e. the host runs a Windows OS and has [Windows container mode enabled][container-mode]).
+
+> **NOTE**: When setting `DOCKER_HOST`, keep in mind that:
+>
+> - any volumes mounted via `pack build --volume <volume> ...` or `docker run --volume <volume> ...` must exist on the _docker host_, not the client machine.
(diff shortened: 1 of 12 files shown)


content/docs/app-developer-guide/build-a-windows-app.md
```diff
@@ -0,0 +1,87 @@
...
+
+```bash
+# clone the repo
+git clone https://github.com/buildpacks/samples
+
+# build the app
+pack build sample-app --path samples/apps/aspnet --builder cnbs/sample-builder:dotnet-framework-1809 --trust-builder
+```
+
+> **TIP:** If you don't want to keep specifying a builder every time you build, you can set it as your default
```
ameyer-pivotal, line 58: In case you're wondering why I changed `Tip` to `TIP` elsewhere in this PR, it felt weird to leave it lowercase here, and it felt weird to not make it consistent elsewhere.

That said, we could also change `NOTE` to `Note` everywhere instead. It may be more consistent that way, as the installation docs do use `Optional:` (vs `OPTIONAL`).

content/docs/app-developer-guide/build-a-windows-app.md
```diff
@@ -0,0 +1,87 @@
++++
+title="Build a Windows app"
+weight=2
+summary="The basics of taking your Windows app from source code to runnable image."
+aliases=[
+    "/docs/using-pack/building-windows-app/"
+]
```
jromero, line 7:
```suggestion
```

Since this is a new page that never existed under `using-pack`, this shouldn't be necessary.

content/docs/app-developer-guide/build-a-windows-app.md
```diff
@@ -0,0 +1,87 @@
...
+
+# build the app
+pack build sample-app --path samples/apps/aspnet --builder cnbs/sample-builder:dotnet-framework-1809 --trust-builder
+```
+
+> **TIP:** If you don't want to keep specifying a builder every time you build, you can set it as your default
+> builder by running `pack set-default-builder <BUILDER>`.
+> <br/><br/>
+> **NOTE:** The requirement to trust Windows builders with `--trust-builder` is temporary and will be no longer
+> be required for `pack` version `0.13.2` and later.
```
jromero, line 62: I'm not sure what this is referring to. Is there an issue associated to this note? 

Would it be a disservice to remove this note, address the issue, and circle back around to remove this flag without needed to explain it or promise a fix?

content/docs/app-developer-guide/build-a-windows-app.md
```diff
@@ -0,0 +1,87 @@
...
+By setting `DOCKER_HOST`, you can use `pack` on any OS to build Windows container images, as long as your remote Docker host is configured to support them (i.e. the host runs a Windows OS and has [Windows container mode enabled][container-mode]).
+
+> **NOTE**: When setting `DOCKER_HOST`, keep in mind that:
+>
+> - any volumes mounted via `pack build --volume <volume> ...` or `docker run --volume <volume> ...` must exist on the _docker host_, not the client machine.
+> - any ports published via `docker run --publish <host-port>:<container-port> ...` will be published on the _docker host_, not the client machine.
+
+### Enable Windows container mode
+
+In order to produce Windows container images, ensure [Windows container mode][container-mode] is enabled in your Docker settings (available only in Docker for Windows). See [Understanding Docker hosts][docker-hosts]) above if you're using a remote Docker host.
```
jromero, line 34:
```suggestion
In order to produce Windows container images, ensure [Windows container mode][container-mode] is enabled in your Docker settings (available only in Docker for Windows). See [Understanding Docker hosts][docker-hosts] above if you're using a remote Docker host.
```

Extra parenthesis `)`.

content/docs/app-developer-guide/build-a-windows-app.md
ameyer-pivotal: Ah I see. Will remove!

content/docs/app-developer-guide/build-a-windows-app.md
ameyer-pivotal: Turns out we just put up a PR to address this, so I'm on board with removing the note.

content/docs/app-developer-guide/build-a-windows-app.md
ameyer-pivotal: Fixed.

ameyer-pivotal: Good catch @jromero!

ameyer-pivotal:
@jromero 

> Also, note that we've added another page recently, you might want to update the weight to co-exist with "Build an app".

Done! Thanks

PR #206, opened 2020-09-01
Add testing of tutorials

jromero:


content/docs/buildpack-author-guide/create-buildpack/build-app.md
```diff
@@ -41,18 +44,17 @@ bundle install
...
+<!-- test:exec -->
 ```bash
 pack build test-ruby-app --path ./ruby-sample-app --buildpack ./ruby-buildpack
 ```
 
 you will see the following output:
 
+The **BUILDING** process...
+
+<!-- test:assert=contains -->
```
yaelharel, line 56: What do you think about keeping the "headers" of the phases:
```
==> DETECTING
...
==> ANALYZING
...
==> RESTORING
...
###The building part that you added
==> RESTORING
...
```

jromero: I can't think of a good programmatic way of skipping certain lines...

Maybe an option for `assert=contains` could be something like `ignore=...` in which it actually replaces `...` with `.*` regex and applies that during matching 🤔. I like the idea in general but I think it would take a bit more work.

jromero: Created this issue to not lose track of this: https://github.com/jromero/ugo/issues/4

PR #217, opened 2020-09-24
Add option to install pack CLI using scoop on Windows

goostleek:
Additional option to install pack CLI (using scoop) on Windows has been documented.
The new way of installation is possible due to ScoopInstaller/Main#1402.

Scoop's packages database is refreshed once a day so it will always install the latest version available.

Signed-off-by: Marcin Kłopotek <marcin.klopotek@gmail.com>

### DIFF
--- a/content/docs/tools/pack/cli/install.md
+++ b/content/docs/tools/pack/cli/install.md
@@ -56,9 +56,13 @@ brew install buildpacks/tap/pack
 ---
 
 ## Windows
-To install `pack` on Windows, we recommend using Chocolatey:
+To install `pack` on Windows, we recommend using [Chocolatey](https://chocolatey.org/):
 ```
 choco install pack --version={{< latest >}}
+```
+or [scoop](https://scoop.sh/):
+```
+scoop install pack
 ```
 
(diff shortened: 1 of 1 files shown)


dfreilich (approved):
Verified that it works; thanks for the PR to scoop to add it! 

I don't know much about how scoop works, but is there any need for us to update the manifest in [scoop](https://github.com/ScoopInstaller/Main), or is it updated automatically?

goostleek:
I'm sorry I think I was not clear enough in the [description](https://github.com/buildpacks/docs/pull/217#issue-492397301).

> Scoop's packages database is refreshed once a day so it will always install the latest version available.
That means each time a new pack CLI [release](https://github.com/buildpacks/pack/releases) appears on GH the manifest will be refreshed by the scoop infrastructure. The database is refreshed on a daily basis so it may take up to a day to have the manifest updated. **There's no need to update the manifest manually in case of a new CLI release**. The version number is updated and checksums are calculated by the update workflow. All other metadata are left intact. However, there are cases you may want to update the manifest (to change licence, repository URL, etc).

jromero (approved):

PR #219, opened 2020-09-29
Document report.toml

natalieparellano:
Still kind of a WIP - it would be nice to have a way to do this with pack. The k8s
solution could use some additional guidance.

### DIFF
--- a/content/docs/app-developer-guide/using-report-toml.md
+++ b/content/docs/app-developer-guide/using-report-toml.md
@@ -1,0 +1,11 @@
++++
+title="Using report.toml"
+weight=7
+summary="Learn how to access the report.toml file to get metadata about a build."
++++
+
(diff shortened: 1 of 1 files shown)


natalieparellano:
We should document this, but this is probably not the right place for it. Looking for a home...

dfreilich:
Has there been any update on where this would go? It would be nice to have some more documentation on the lifecycle, perhaps under https://buildpacks.io/docs/tools/ (lifecycle?)

natalieparellano:
> Has there been any update on where this would go? It would be nice to have some more documentation on the lifecycle, perhaps under https://buildpacks.io/docs/tools/ (lifecycle?)

Yes! The implementation team has been discussing adding more docs for the lifecycle. Look out for some new issues...

PR #226, opened 2020-10-06
Compact installation instructions into tab format

jromero:
Compact installation instructions into tab format

Also,

- Move installation instruction to be front and center for pack
- Add pack as a Go library information

### Before

buildpacks io_docs_tools_pack_cli_install_

### After
localhost_1313_docs_tools_pack_


data/pack/install.yml
```diff
@@ -0,0 +1,79 @@
+- name: Linux
+  methods:
+    - name: Arch Linux
+      content: |
+        On Arch Linux, there are two packages available:
+        
+        - [pack-cli](https://aur.archlinux.org/packages/pack-cli/)
+        - [pack-cli-bin](https://aur.archlinux.org/packages/pack-cli-bin/)
```
dfreilich, line 8: How do you download from Arch Linux? Any CLI commands?

data/pack/install.yml
```diff
@@ -0,0 +1,79 @@
...
+    - name: Command
```
dfreilich, line 9:
```suggestion
    - name: Script Install
```

data/pack/install.yml
```diff
@@ -0,0 +1,79 @@
...
+      content: |
+        The following command will download the latest version of `pack` from [GitHub](https://github.com/buildpacks/pack/releases) and install it in `/usr/local/bin/`.
+
+        ```bash
+        (curl -sSL "https://github.com/buildpacks/pack/releases/download/v<PACK_VERSION>/pack-v<PACK_VERSION>-linux.tgz" | sudo tar -C /usr/local/bin/ --no-same-owner -xzv pack)
+        ```
+    - &container
+      name: Container
+      content: |
+        `pack` is also available as container images on Docker Hub as [`buildpacksio/pack`](https://hub.docker.com/r/buildpacksio/pack)
```
dfreilich, line 19:
```suggestion
        `pack` is available as a container image on Docker Hub as [`buildpacksio/pack`](https://hub.docker.com/r/buildpacksio/pack)
```

data/pack/install.yml
```diff
@@ -0,0 +1,79 @@
...
+      content: |
+        To manually install `pack`, visit the [releases page on GitHub](https://github.com/buildpacks/pack/releases). 
+        1. Download the `.tgz` or `.zip` file for your platform
+        2. Extract the `pack` binary
+        3. (Optional) Add the directory containing `pack` to `PATH`
+- name: macOS
+  default: true
+  methods:
+    - *container
+    - *homebrew
```
dfreilich, line 62:
```suggestion
    - *homebrew
    - *container
```

data/pack/install.yml
```diff
@@ -0,0 +1,79 @@
...
+      default: true
+      content: |
+        To install `pack` on Windows, we recommend using [Chocolatey](https://chocolatey.org/):
+        ```
+        choco install pack --version=<PACK_VERSION>
+        ```
+    - *container
+    - name: Scoop
+      content: |
+        ```
```
dfreilich, line 76:
```suggestion
       `pack` can be installed via [Scoop](https://scoop.sh/):
        ```
```

data/pack/install.yml
```diff
@@ -0,0 +1,79 @@
...
+        1. Download the `.tgz` or `.zip` file for your platform
+        2. Extract the `pack` binary
+        3. (Optional) Add the directory containing `pack` to `PATH`
+- name: macOS
+  default: true
+  methods:
+    - *container
+    - *homebrew
+    - *manually
+- name: Windows
```
dfreilich, line 64: I would order it as:
```
* Chocolatey
* Scoop
* Container
* Manually
```

So that both CLI install methods are next to each other (because anyways, the way it is now, it's not alphabetical). What do you think?

dfreilich (changes requested):
This looks super super cool. I really like this change

data/pack/install.yml
jromero: For AURs, it's dependant on the package manager they use so we opted to be non-prescriptive.

data/pack/install.yml
jromero: I choose to keep the tabs alphabetically ordered as to not be bias towards any tool although we kind of are when we default to "homebrew". Open to discussion but I prefer alphabetical when looking at various options.

PR #237, opened 2020-10-26
Add migration guide for platform API 0.3 -> 0.4

### LINKED ISSUE #233: Migration guide: platform API 0.3 -> 0.4
## Summary

We should publish a migration guide for buildpacks users wishing to upgrade from platform API 0.3 -> 0.4

* What are the changes?
* Who are the changes relevant for? (end users, platform operators, etc.)
* What actions need to be taken?
* Links to other relevant materials (e.g., Tekton task implementing this platform API version)

- [ ] App Developer
- [ ] Buildpack Author
- [x] Operator (but could be others)
- [ ] Other:

## Additional Information

We can work off of the information in the [spec release notes](https://github.com/buildpacks/spec/releases/tag/platform%2Fv0.4) and this [Google doc](https://docs.google.com/document/d/1BYQSFcQHOgbIjx6c6RzcsvHPJyQwyd6TQhIYBy48cRw/edit)

natalieparellano:
Fixes #233 


content/docs/reference/spec/migration/platform-api.md
```diff
@@ -0,0 +1,42 @@
...
+
+Windows image builds are now supported! See the [Windows build guide](/docs/app-developer-guide/build-a-windows-app)
+for more details.
+
+### Multicall launcher
+
+When running the exported application image, it is now possible to provide arguments to pre-defined process types. 
+The new flow is described [here](/docs/app-developer-guide/run-an-app/#run-a-multi-process-app).
+As part of these changes, the exporter will not set, and the launcher no longer accepts, the `CNB_PROCESS_TYPE` variable.
+See the associated [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0045-launcher-arguments.md) and [spec PR](https://github.com/buildpacks/spec/pull/118)
```
yaelharel, line 22: Personally I think that adding links to the RFC and spec PR is too much for this type of document, but I think we should allow the reader to get more information about the new feature in a different way.
What do you think about adding another page or section in this page with more details about each change in the new API?

yaelharel (commented):
Personally I think that adding links to RFCs and spec PRs is too much for this type of document, but I think we should allow the reader to get more information about the new feature in a different way.
What do you think about adding another page or section in this page with more details about each change in the new API? We can also add an example of the usage of each one of them.

natalieparellano:
> Personally I think that adding links to RFCs and spec PRs is too much for this type of document, but I think we should allow the reader to get more information about the new feature in a different way.
> What do you think about adding another page or section in this page with more details about each change in the new API? We can also add an example of the usage of each one of them.

I agree. Personally I think it's best if we link to other sections of the docs (like we do for the launcher in this PR). But sometimes, those docs don't exist yet (like for the BOM). At a minimum when we create these guides we should ensure that issues to create the other docs exist.

content/docs/reference/spec/migration/platform-api.md
```diff
@@ -0,0 +1,42 @@
++++
+title="Platform API 0.3 -> 0.4"
```
jromero, line 2: The title of this doesn't match the file name. I think we'd like to prepare ourselves for having multiple migration guides for platform API hence I would recommend that the file name be changed to `platform-api-0.3-0.4.md`

jromero:
> > Personally I think that adding links to RFCs and spec PRs is too much for this type of document, but I think we should allow the reader to get more information about the new feature in a different way.
> > What do you think about adding another page or section in this page with more details about each change in the new API? We can also add an example of the usage of each one of them.
> 
> I agree. Personally, I think it's best if we link to other sections of the docs (like we do for the launcher in this PR). But sometimes, those docs don't exist yet (like for the BOM). At a minimum when we create these guides we should ensure that issues to create the other docs exist.

I agree with the idea. I have concerns about the overhead. I think creating issues at a minimum still isn't much better if they aren't available or coordinated. 

In a perfect world, all related docs, at least minimally, are merged at once. This would increase the amount of effort necessary for a migration guide. I think referencing PRs and RFCs is extremely valuable but it shouldn't be used as a way to "learn" about a change. Instead, it should be seen as helpful method to trace down the _reasoning_ for a change.

jromero (approved):
@natalieparellano this is great work! I'm more than happy merging this in until we find a better alternative to pointing at PRs.

PR #240, opened 2020-10-28
Add link check

natalieparellano:


dfreilich (approved):
Assuming we can get it passing...

woohoo

natalieparellano:
Example of failing check: https://github.com/buildpacks/docs/pull/240/checks?check_run_id=1333056573

Note that we're only looking for "does not exist" errors. There are also:
* alt attribute missing
* link tag missing href
* target is a directory, href lacks trailing slash

I decided to ignore those for now (there are quite a few of them). We may wish to fix them in the future.

PR #245, opened 2020-11-09
Clarify builder definition

abitrolly:
Single builder cares for one app only, right?
`lifecycle` is a program, which has an implementation.

What is still unclear for me is how "`platforms` use" environment. For me `platform` is always an abstract term. Just a comment.

### DIFF
--- a/content/docs/concepts/components/builder.md
+++ b/content/docs/concepts/components/builder.md
@@ -1,7 +1,7 @@
 +++
 title="Builder"
 weight=1
-summary="A builder is an image that bundles all the bits and information on how to build your apps, such as buildpacks, an implementation of the lifecycle, and a build-time environment that platforms may use when executing the lifecycle."
+summary="A builder is an image that bundles all the bits and information on how to build your app. It contains buildpacks, an implementation of the lifecycle program, and a build-time environment that platforms may use when executing the lifecycle."
 aliases=[
     "/docs/using-pack/working-with-builders/"
 ]


dfreilich:
@abitrolly Thanks for the PR and feedback! To your last point – does our [platform](https://buildpacks.io/docs/concepts/components/platform/) documentation clarify that at all? If not, I'd love to brainstorm on how to make it a bit more of a tangible concept.

abitrolly:
@dfreilich well, I tried to edit more docs to clarify, but DCO is painful for editing from GitHub https://github.com/probot/probot/issues/1403

I would link `platform` to the documentation you've mentioned. If such links work for the `summary`.

In particular it is surprising to read that `pack` is a platform. I would say that `pack` is a tool that creates and maintains "build environment" for running builders. If that is correct..

dfreilich:
Unfortunately, looks like links can't be part of a summary as it is. If you'd like, you can pull the summary into the main text of the file, and add a link to it within it, like so:

```
A builder is an image that bundles all the bits and information on how to build your app. It contains buildpacks, an
implementation of the lifecycle program, and a build-time environment that [platforms][platforms] may use when executing
the lifecycle.
<!--more-->
...
[platforms]: /docs/concepts/components/platform
```

Re: pack as a platform:
`pack` is a platform, insofar as it facilitates transformations to the application by applying the builder/lifecycle to it. It isn't the most intuitive use of `platform`, but the other platforms (such as `tekton`, `circleCI` or `kpack`) are definitely more intuitive. Does that make sense?

abitrolly:
I don't like that `platform/builder/lifecycle` abstraction at all.

> pack is a platform, insofar as it facilitates transformations to the application by applying the builder/lifecycle to it. 

`pack` is a program that executes `builder` in phases to create a container from an application source

> It isn't the most intuitive use of platform, but the other platforms (such as tekton, circleCI or kpack) are definitely more intuitive.

`tekton`, `cirrcleCI` or `kpack` also provide online services that execute  `builder` for you. I haven't looked into them, so I am not sure how is that correct.

abitrolly:
As for the links in summary, mayhe Hugo has an open issue about that. I am not sure. But referencing content for inclusion from other pages looks reasonable/

jromero:
A platform is an orchestrator of the lifecycle phases along with the source code of the application. By that generic definition, Pack (as well as Tekton, kpack, CircleCI, Heroku, etc) are platforms. 

Does this diagram provide some clarity? 
image

Here are a few more examples of existing platforms:
image

content/docs/concepts/components/builder.md
```diff
@@ -1,7 +1,7 @@
 +++
 title="Builder"
 weight=1
-summary="A builder is an image that bundles all the bits and information on how to build your apps, such as buildpacks, an implementation of the lifecycle, and a build-time environment that platforms may use when executing the lifecycle."
+summary="A builder is an image that bundles all the bits and information on how to build your app. It contains buildpacks, an implementation of the lifecycle program, and a build-time environment that platforms may use when executing the lifecycle."
```
jromero, line 4: A builder can most certainly build multiple applications. Not at once of course but the same builder can be used to build java apps, node apps, ruby apps, etc. For example, you can see all the language supports on the [Google builder](https://github.com/GoogleCloudPlatform/buildpacks/blob/main/builders/gcp/base/builder.toml#L103
).
 
```suggestion
summary="A builder is an image that bundles all the bits and information on how to build your apps. It contains buildpacks, an implementation of the lifecycle program, and a build-time environment that platforms may use when executing the lifecycle."
```

abitrolly: How about adding it with the separate sentence "The same builder can build apps in different languages. When analysing code, it automatically selects appropriate buildpack for the target language." ?

abitrolly:
If I me from a few months ago asked my about the builder now, the dialog would be like this.

> So, what is a builder?
A builder is an image.

> You mean Docker image.
Yep.

> So what does it do?
It builds apps.

> So it builds app in container, so you can get your binaries without installing all those SDKs and such.
Not really. It builds another image, so you can immediately run your binaries with all dependencies and stuff.

> But that's what Dockerfile is for.
You need to write Dockerfile for every app, and builder just bullds. Automatically.

> Even I don't know how to build our app, and you say it can figure it out. :D
Well, you need to follow some best practices of course.

> And if I don't?
Then you need to create your own buildpack and tell builder to use it.

> What's this buildpack?
Another image.

> Do you mean I need my custom builder image?
No. You need custom buildpack image.

> I don't get it. What is buildpack?
Builder knows nothing about your project.

> Now we're talking.. So..
It uses buildpacks to build your files.

> And why I can not just run that buildpack directly?
You can.

> So.. Why I need this builder?
It selects the buildpack for you.

> You said it knows nothing about me project.
That's why it runs buildpacks one after another, and they tell if they can build the project.

> Sound slow. Where does this builder get those build-packs?
The builder image contains those buildpacks.

> Whaaat? You said the builder is an image.
Yep.

> The image that contains another image?
That's correct.

> You run container that runs another container inside itself? No way.
Looks like how it is.

> That's should be horribly slow and that builder image must be huge.
I wish I could tell you exact size, but I don't know.

> I still don't get how a container can access another image inside itself to run another container.
Me too.

> Okay. So I need to build a custom buildpack, then my own builder.. Why I won't just use Dockerfile?
Use Dockerfile.

> Your builder is useless.
Yep. But it make sense if you have multiple similar apps, or want to use a single way to build them all.

> Nah. I need some better reasons.
I guess so. :D

content/docs/concepts/components/builder.md
jromero: I'm wary about using singular _app_ because I would hate for the initial perception to be that there's a one-to-one relationship between app and builder. What are your concerns with a plural?

In regards to the additional blurb about languages I would say:

1. I think it would be great to add this information but I wouldn't put it in the summary. I would add another paragraph talking about the one-to-many potentiality.
2. The idea of a single builder supporting multiple languages is just an example and when added it should be noted as such. Another example could be that you can simple have a javascript builder (if you are a strict javascript shop) that understands various development stacks (ie. node with yarn vs node with npm, etc).


Maybe something like (a _very rough_ draft):

> Builders are typically not created specifically for one application. Instead, they are composed of multiple buildpacks that when the code is analyzed the proper buildpacks are operate against it.
>
> For example, the same builder can build apps in different languages. When analysing code, it would automatically select the appropriate buildpacks for the target language.
>

jromero:
I ❤️  this honest hypothetical dialog. I can see where you are coming from. I'm sure there's still a lot more to discuss here. Are you in our [slack](https://slack.buildpacks.io)? It would be great if we could get some more direct dialog going.

content/docs/concepts/components/builder.md
abitrolly: My concerns with plural as a DevOps is that the relationship between a builder and produced artifact is one-to-one. If I need to get multiple artifacts (back, front and worker), I need to run the builder multiple times with multiple targets. Compare this to Kubernetes - I run it once and it executes multiple artifacts. That's how one-to-many relationship works for me.

abitrolly: > Another example could be that you can simple have a javascript builder (if you are a strict javascript shop) that understands various development stacks (ie. node with yarn vs node with npm, etc).

I thought that builder can't understand anything. It is the role of buildpack to analyse the project and tell if the buildpack can build it, no?

> Builders are typically not created specifically for one application. Instead, they are composed of multiple buildpacks that when the code is analyzed the proper buildpacks are operate against it.

This leaves too many open questions. Why builders are composed of buildpacks? Who and when analyses the code? If buildpacks operate against code on their own, why builder is needed at all?

abitrolly:
@jromero Slack multiplies chat fatigue by password fatigue. ) I used it when I had a job, and it ate a lot of memory, so no slacking anymore. )

sampeinado:
Hey @abitrolly! I am working on getting some [feedback from users](https://github.com/sampeinado/rfcs/blob/main/text/0000-CNB-user-research-2021.md) about the buildpacks UX and what's important. Would you be up for a 60-90 min interview this week or next so that I can learn more from your experience? We would compensate $250 for your time!

abitrolly:
Hi @sampeinado. No problem, I am usually available 11:00-13:00 and 16:00-22:00 UTC+3 each day. You can use anatoli@rainforce.org to schedule a meeting.

jkutner:
Superseded by #309

PR #260, opened 2020-11-19
Promoted to incubation from sandbox

jkutner:
https://www.cncf.io/blog/2020/11/18/toc-approves-cloud-native-buildpacks-from-sandbox-to-incubation/


dfreilich:
Could we also add a banner to the homepage that we were promoted?

jromero (approved):

jromero:
> Could we also add a banner to the homepage that we were promoted?
> 
> huge

Do it!

PR #265, opened 2020-11-30
Added description of Buildpack URIs

jkutner:
Updated "Specify buildpacks" with details from [RFC-0037](https://github.com/buildpacks/rfcs/blob/main/text/0037-buildpack-uris.md)

### DIFF
--- a/content/docs/app-developer-guide/specific-buildpacks.md
+++ b/content/docs/app-developer-guide/specific-buildpacks.md
@@ -6,7 +6,7 @@ summary="Learn how to specify exactly what buildpacks are used during the build 
 
 You may specify exactly what buildpacks are used during the build process by referencing them with a URI in any of the following formats.
 
-| Name | Format |
+| Type | Format |
 |----- |------- |
 | Relative | `<path>` |
 | Filesystem | `file://[<host>]/<path>` |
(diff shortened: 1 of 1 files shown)


content/docs/app-developer-guide/specific-buildpacks.md
```diff
@@ -4,19 +4,29 @@ weight=4
 summary="Learn how to specify exactly what buildpacks are used during the build process."
 +++
 
-{{< param "summary" >}}
+You may specify exactly what buildpacks are used during the build process by referencing them with a URI in any of the following formats.
 
-### `--buildpacks`
+| Type | Format |
```
jromero, line 9: I was going to argue that this was Pack specific and should be noted as such but given that we want support for buildpack URIs in the project descriptor which are portable it makes me wonder if buildpack URIs should be spec'd since they we will be more endemic than we originally thought. 

I can see them as part of the [distribution](https://github.com/buildpacks/spec/blob/main/distribution.md) spec but maybe an extension spec is more appropriate.

Thoughts?

CC: @elbandito

jromero (approved):
This is great! I left one forward looking comment. Not expecting to resolve the thread here unless it changes the outcome but probably something worth thinking about.

content/docs/app-developer-guide/specific-buildpacks.md
jkutner: Yea, I agree. I'll draft an extension spec, and we can decide if it should move into distribution (which I agree makes the most sense)

PR #267, opened 2020-12-03
Add community page with calendar

jromero:
## Preview

localhost_1313_community_


content/community.md
```diff
@@ -0,0 +1,12 @@
++++
+title="Community"
+weight=2
+layout="community"
+summary="Cloud Native Buildpacks provide a unique solution to the image build process. See how it compares to the alternatives."
```
dfreilich, line 5: nit: Remove this?

dfreilich (changes requested):
This looks super super cool, thanks for adding it! 

Some specific things I think could be clarified/changed: 
* Clarify the time zone, because I was very excited when I thought that the syncs were now going to be in 10am Israel time 😄 
* I was a bit confused that clicking on the dates took me to the doc, and not like a Google/Outlook invite. Maybe that's intended, but then clarify what clicking on the links will do?
* Is there any way to backfill dates? Anything before this week is empty. 
* Maybe clarify that the "Add these events to your calendar" is an .ics file, and not "importing into your Google Calendar/other calendar"? Alternatively, if there could be an "import into your calendar", that would also be great
* Maybe also add Github to the "Join our community" list?

jromero:
I'm going to preface this with the idea that I'm trying iterate over this so some of my answer most likely be geared towards that.

---

> * Clarify the time zone, because I was very excited when I thought that the syncs were now going to be in 10am Israel time 😄

For now, I'm going to add a disclaimer that mentions that all times are in EST/EDT. The JS standard date/time library isn't the greatest so it would take a bit of effort to properly adjust to local timezone.

> * I was a bit confused that clicking on the dates took me to the doc, and not like a Google/Outlook invite. Maybe that's intended, but then clarify what clicking on the links will do?

The eventual hope is to produce a modal/popup that provides the following:

- Name
- Description
- Import into your calendar

> * Is there any way to backfill dates? Anything before this week is empty.

Yes, how far back do you want to go? I wasn't expecting a typical user to look back in time.

> * Maybe clarify that the "Add these events to your calendar" is an .ics file, and not "importing into your Google Calendar/other calendar"? Alternatively, if there could be an "import into your calendar", that would also be great

Will add some attribution in the short-term. 

Long-term I'd like to do something similar to what GitHub Universe conf did...

image

image


> * Maybe also add Github to the "Join our community" list?

I thought about this but wasn't sure what we'd link them to that is "community". I'd hate to simply dump them in the org page with a list of all the repos. Additionally, that link is already at the top right of the page. Maybe the [community repo](https://github.com/buildpacks/community/)?

jromero:
@dfreilich was able to make a few changes:

1. Show the events in local time (see if it works well for you)
2. Added community GitHub repo to list
3. Updated the wording for `.ics` files

## Preview

localhost_1313_community_ (1)

jromero:
Issues created:

- https://github.com/buildpacks/docs/issues/268
- https://github.com/buildpacks/docs/issues/269

PR #289, opened 2021-01-21
Add docs for using the cache-image flag

dwillist:
Signed-off-by: dwillist <dthornton@vmware.com>

This PR adds documentation for the `cache-image` flag. It should likely be expanded on with a section about how pack uses cache volumes & one about using cache images in CI. 

The PR does not use `ugo` to test the commands as they require docker hub authentication (which is difficult to automate).

### DIFF
--- a/content/docs/app-developer-guide/using-cache-image.md
+++ b/content/docs/app-developer-guide/using-cache-image.md
@@ -1,0 +1,77 @@
++++
+title="Cache Images"
+weight=4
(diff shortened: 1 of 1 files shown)


dfreilich:
That alone is great, but I'm excited to see the extensions to this docs! This may be a valuable starting point, though, and perhaps we should have links between this and https://buildpacks.io/docs/app-developer-guide/mouting-volumes/?

jromero (approved):

PR #291, opened 2021-01-27
layer.toml's metadata is a table not a string in the spec

hone:
I was going through the docs yesterday and noticed that we're telling people to treat the `metadata` as a string value where the [spec uses a table](https://github.com/buildpacks/spec/blob/main/buildpack.md#layer-content-metadata-toml).


danielleadams (approved):

jromero:
Good catch @hone. It looks like we use something similar in the [following page](https://github.com/buildpacks/docs/blob/main/content/docs/buildpack-author-guide/create-buildpack/make-buildpack-configurable.md). Do you mind updating that reference as well?

hone:
@jromero yep, I can fix that.

jkutner (approved):

samj1912:
It looks like some of the references were not updated. I tried to address that here https://github.com/buildpacks/docs/pull/300

PR #297, opened 2021-02-09
Reworked the composition of build-app step in Buildpack author guide

jkutner:
I've talked with a few people who've asked "how do I create layers in my buildpack". I don't think the existing docs split this part out enough to make it understandable (and the current doc isn't something that's good to directly link to if you want answer that question).

### DIFF
--- a/content/docs/buildpack-author-guide/create-buildpack/build-app.md
+++ b/content/docs/buildpack-author-guide/create-buildpack/build-app.md
@@ -20,6 +20,62 @@ Let's begin by changing the `ruby-buildpack/bin/build` so that it creates a laye
 
 A Buildpack layer is represented by a directory inside the [layers directory][layers-dir] provided to our buildpack by the Buildpack execution environment. To create a new layer directory representing the Ruby runtime, change the `build` script to look like this:
 
+```bash
+#!/usr/bin/env bash
+set -eo pipefail
+
+echo "---> Ruby Buildpack"
+
+layersdir=$1
+
+rubylayer="$layersdir"/ruby
+mkdir -p "$rubylayer"
+```
+
+The `rubylayer` directory is a sub-directory of the directory provided as the first positional argument to the script (the [layers directory][layers-dir]), and this is where we'll store the Ruby runtime.
+
+Next, we'll download the Ruby runtime and install it into the layer directory. Add the following code to the end of the `build` script:
+
+```bash
+echo "---> Downloading and extracting Ruby"
+ruby_url=https://s3-external-1.amazonaws.com/heroku-buildpack-ruby/heroku-18/ruby-2.5.1.tgz
+wget -q -O - "$ruby_url" | tar -xzf - -C "$rubylayer"
+```
+
+This code uses the `wget` tool to download the Ruby binaries from the given URL, and extracts it to the `rubylayer` directory.
+
+The last step in creating a layer is writing a TOML file that contains metadata about the layer. Without this file, the Buildpack lifecycle will ignore the layer directory. For the Ruby layer, we need to ensure it is available in the launch image by setting the `launch` key to `true`. Add the following code to script:
+
+```bash
+echo -e 'launch = true' > "$rubylayer.toml"
+```
+
+### Installing Dependencies
+
+Next, we'll use the Ruby runtime you installed to download the application's dependencies. First, we need to make the Ruby executables available to our script by putting it on the Path. Add the following code to the end of the `build` script:
+
+```bash
+export PATH="$rubylayer"/bin:$PATH
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+${LD_LIBRARY_PATH}:}"$rubylayer/lib"
+```
+
+Now we can install Bundler, a dependency manager for Ruby, and run the `bundle install` command. Append the following code to the script:
+
+```bash
+echo "---> Installing bundler"
+gem install bundler --no-ri --no-rdoc
+
+echo "---> Installing gems"
+bundle install
+```
+
+Now the Buildpack is ready to test.
+
+### Running the Build
+
+Your complete `build` script should look like this:
+
+
 <!-- test:file=ruby-buildpack/bin/build -->
 ```bash
 #!/usr/bin/env bash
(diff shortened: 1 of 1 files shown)


content/docs/buildpack-author-guide/create-buildpack/build-app.md
```diff
@@ -19,20 +30,44 @@ echo "---> Ruby Buildpack"
...
+The last step in creating a layer is writing a TOML file that contains metadata about the layer. Without this file, the Buildpack lifecycle will ignore the layer directory. For the Ruby layer, we need to ensure it is available in the launch image by setting the `launch` key to `true`. Add the following code to script:
+
+```bash
+# 4. MAKE RUBY AVAILABLE DURING LAUNCH
 echo -e 'launch = true' > "$rubylayer.toml"
+```
+
+### Installing Dependencies
+
+Next, we'll use the Ruby runtime you installed to download the applications dependencies. First, we need to make the Ruby executables available to our script by putting it on the Path. Add the following code to the end of the `build` script:
```
natalieparellano, line 60:
```suggestion
Next, we'll use the Ruby runtime you installed to download the application's dependencies. First, we need to make the Ruby executables available to our script by putting it on the Path. Add the following code to the end of the `build` script:
```

content/docs/buildpack-author-guide/create-buildpack/build-app.md
```diff
@@ -19,20 +30,44 @@ echo "---> Ruby Buildpack"
...
 
+```bash
 # 4. MAKE RUBY AVAILABLE TO THIS SCRIPT
 export PATH="$rubylayer"/bin:$PATH
 export LD_LIBRARY_PATH=${LD_LIBRARY_PATH:+${LD_LIBRARY_PATH}:}"$rubylayer/lib"
+```
+
+Now we can install Bundler, a dependency manager for Ruby, and run the `bundle install` command. Append the following code to the script:
 
+```bash
```
natalieparellano, line 70: Maybe it's worth calling out where the result of these commands end up - i.e., in the ruby directory and/or workspace. 

This is a bit more involved, but I feel like this tutorial could benefit from a `dive test-ruby-app` mapping each of the commands to a layer in the final image.

content/docs/buildpack-author-guide/create-buildpack/build-app.md
```diff
@@ -5,9 +5,20 @@ weight=404
...
+
+By doing this, you'll learn how to create arbitrary layers with your buildpack, and how to read the contents of the app in order to perform actions like downloading dependencies.
+
+Let's begin by changing the `ruby-buildpack/bin/build` so that it creates a layer for Ruby.
+
+### Creating a Layer
+
+A Buildpack layer is represented by a directory inside the [layers directory][layers-dir] provided to our buildpack by the Buildpack execution environment. To create a new layer directory representing the Ruby runtime, change the `build` script to look like this:
 
 <!-- test:file=ruby-buildpack/bin/build -->
```
jromero, line 23: I would think that the tests are failing because the file being created for the test is now incomplete since it's been broken down into multiple sections. [ugo](https://github.com/jromero/ugo#usage), doesn't have support for appending to a file. It would be a great feature to add. :)

jkutner: For now, I could add a file step to the tutorial that says "the full file should look like this"

content/docs/buildpack-author-guide/create-buildpack/build-app.md
jkutner: I'm not sure I follow, but if that's an enhancement to `build` script, I'd rather put that in a separate PR. I was mostly hoping to reorganize this so it explains each step (and has an anchor link for creating a layer)

jromero: What I understood from @natalieparellano's comment was that it may be helpful to add a visual on how each set of commands translate to layers on the final image.

ie. similar to what dive displays

image

samj1912:
Another thing re:layers that tripped me up when I was first following this guide was how caching works and that you need to clean up all the files from the previous build if you don't want any stale files. I added a comment about it at https://github.com/buildpacks/docs/pull/300#discussion_r577282797

but I think it probably deserves a bigger section. Libraries like packit provide a convenient reset method for a layer. https://github.com/paketo-buildpacks/packit/blob/ee8f52e652f9a349ca47b3e709289d954b8de6a5/layer.go#L64

In the gem tutorial we currently keep previous gems around even if they are not needed/used. This can be a sort of quirky concept to wrap your head around as a BP author and something that a beginner should be aware of.

jromero (approved):

content/docs/buildpack-author-guide/create-buildpack/build-app.md
```diff
@@ -5,9 +5,76 @@ weight=404
...
+```
+
+Now we can install Bundler, a dependency manager for Ruby, and run the `bundle install` command. Append the following code to the script:
+
+```bash
+echo "---> Installing bundler"
+gem install bundler --no-ri --no-rdoc
+
+echo "---> Installing gems"
+bundle install
```
dfreilich, line 69: Is there any benefit to doing this in a separate layer?

content/docs/buildpack-author-guide/create-buildpack/build-app.md
```diff
@@ -5,9 +5,76 @@ weight=404
...
+ruby_url=https://s3-external-1.amazonaws.com/heroku-buildpack-ruby/heroku-18/ruby-2.5.1.tgz
+wget -q -O - "$ruby_url" | tar -xzf - -C "$rubylayer"
+```
+
+This code uses the `wget` tool to download the Ruby binaries from the given URL, and extracts it to the `rubylayer` directory.
+
+The last step in creating a layer is writing a TOML file that contains metadata about the layer. Without this file, the Buildpack lifecycle will ignore the layer directory. For the Ruby layer, we need to ensure it is available in the launch image by setting the `launch` key to `true`. Add the following code to script:
+
+```bash
+echo -e 'launch = true' > "$rubylayer.toml"
```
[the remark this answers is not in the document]
samj1912: I agree. I also got tripped up by this when I first read the tutorial.

content/docs/buildpack-author-guide/create-buildpack/build-app.md
samj1912: I think it's useful in the long term if someone wants to implement caching for the ruby distribution. (Currently it's left as an exercise for users at the end of the tutorial). Maybe we can foreshadow it a bit or say that it helps to organize BP layers like this in case you want better control over different kinds of dependencies.

content/docs/buildpack-author-guide/create-buildpack/build-app.md
```diff
@@ -5,9 +5,76 @@ weight=404
...
+mkdir -p "$rubylayer"
+```
+
+The `rubylayer` directory is a sub-directory of the directory provided as the first positional argument to the script (the [layers directory][layers-dir]), and this is where we'll store the Ruby runtime.
+
+Next, we'll download the Ruby runtime and install it into the layer directory. Add the following code to the end of the `build` script:
+
+```bash
+echo "---> Downloading and extracting Ruby"
+ruby_url=https://s3-external-1.amazonaws.com/heroku-buildpack-ruby/heroku-18/ruby-2.5.1.tgz
```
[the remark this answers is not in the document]
jkutner: No, because we're still on bionic (Ubuntu 18.04) in this tutorial

content/docs/buildpack-author-guide/create-buildpack/build-app.md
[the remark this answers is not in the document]
jkutner: Addressed in c38a0bc

content/docs/buildpack-author-guide/create-buildpack/build-app.md
jkutner: I think this would good for a future PR

content/docs/buildpack-author-guide/create-buildpack/build-app.md
```diff
@@ -5,9 +5,76 @@ weight=404
...
+wget -q -O - "$ruby_url" | tar -xzf - -C "$rubylayer"
+```
+
+This code uses the `wget` tool to download the Ruby binaries from the given URL, and extracts it to the `rubylayer` directory.
+
+The last step in creating a layer is writing a TOML file that contains metadata about the layer. Without this file, the Buildpack lifecycle will ignore the layer directory. For the Ruby layer, we need to ensure it is available in the launch image by setting the `launch` key to `true`. Add the following code to the build script:
```
samj1912, line 47: Is it worth adding a note that the toml name must match the layer name it is associated with?

jkutner: Fixed in 6da86e3

PR #310, opened 2021-02-24
Pin workflow (CI) to use go 1.15

jromero:
This is a temporary solution to unblock our pipelines but we'd want to figure out how to successfully upgrade to work with go1.16.

Context:
- [Failing build](https://github.com/buildpacks/docs/runs/1973682785?check_suite_focus=true)


jromero:
Created an issue to capture go1.16 upgrade: https://github.com/buildpacks/docs/issues/311

PR #69, opened 2019-10-08
Restructure menu and add concepts section

jromero:
* Use samples repo for examples
* Add bootstrap

Signed-off-by: Javier Romero <jromero@pivotal.io>

A few screenshots:

image

image

PR #334, opened 2021-03-23
Fix broken link to builder configuration file

edmorley:
Since as of #69, the "Builder Configuration" section no longer exists on:
https://buildpacks.io/docs/concepts/components/builder/

The link now targets the reference section instead:
https://buildpacks.io/docs/reference/config/builder-config/


jromero (approved):
🙏 Thank you!

PR #337, opened 2021-03-23
Update the "current" API version in the buildpack API reference

edmorley:
Since Buildpack API version `0.5` is the latest, not `0.2`:
https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.5


jromero:
@buildpacks/implementation-maintainers could you review the remainder contents of this document to ensure that it does align with Buildpack API 0.5?

jkutner (approved):

edmorley:
I guess this is out of date now that buildpack API 0.6 has been released?

jkutner:
@edmorley yea, still worth getting it in I think. sorry we let this slip, but thanks for fixing.

PR #368, opened 2021-05-18
Add dockerfile to the feature comparison list

### LINKED ISSUE #355: Add Dockerfile comparison
As a user, I would like to see a [comparisons](https://buildpacks.io/features/#comparison) against `docker build` (aka Dockerfiles).

samj1912:
First take at comparing dockerfiles to buildpacks.

I might need some help with aligning the grid. Couldn't figure out how to make the feature elements `col-sm-6` or show up 2 at a time so that they are even.

Based on the learning sync today, we should also add an addition section under "reference" that digs deeper into Source -> Image solutions vs something like Dockerfile.

Fixes #355


jkutner (approved):

samj1912:
Added a target image type as well - https://github.com/buildpacks/docs/pull/368/commits/da37b95b41c7f34ae22e7432760d2c37ead55e58

Not sure if there is a better way to phrase this.

samj1912:
image

samj1912:
> I'm fine with the "Target image" row, but I think it might be something there's less consensus on. Buildpacks are used from some strange things like testing mobile apps. And I'm not sure I can say if ko is only for apps.

I didn't know the right phrasing for target image. I wanted to convey that the end image created by buildpacks is generally intended for end-user applications as opposed to base images or os container images that you may use as building blocks for other images. 🤔 

Re ko: https://github.com/google/ko#ko-easy-go-containers I went off of `ko is a simple, fast container image builder for Go applications.` but I didn't dig too deep.

jkutner:
@samj1912 yea, i think it's fine as-is. was just adding my 2-cents :)

themes/buildpacks/layouts/shortcodes/comparison-table.html
```diff
@@ -154,16 +193,26 @@
     <tr>
       <th class="p-3">Governance</th>
       <td class="border-left border-right">CNCF</td>
+      <td>Docker</td>
       <td>Red Hat</td>
       <td>Google</td>
       <td>Google</td>
     </tr>
+    <tr class="border-top bg-light">
+      <th class="p-3">Target image types</th>
```
jromero, line 202:
```suggestion
      <th class="p-3">Best for Building...</th>
```

Not sure this is a better title but at minimum we should use Title Case.

themes/buildpacks/layouts/shortcodes/comparison-table.html
```diff
@@ -154,16 +193,26 @@
...
+      <td class="border-left border-right">Applications</td>
+      <td class="">Applications, Base images, Operating Systems</td>
```
jromero, line 204:
```suggestion
      <ul class="list-circle">
          <li>Applications</li>
          <li>Base Images</li>
          <li>Operating Systems</li>
        </ul>
```

Given it's a list I think it would look better formatted as such.

themes/buildpacks/layouts/shortcodes/comparison-table.html
```diff
@@ -154,16 +193,26 @@
...
+    </tr>
     </tbody>
   </table>
 </div>
 <div class="py-4 small">
   <ul class="list-unstyled">
     <li><sup>&dagger;</sup> This feature does not apply to solution.</li>
+    <li><sup>&diams;</sup> Requires additional configuration.</li>
     <li><sup>&ddagger;</sup> Requires additional configuration which adversely affects other features.</li>
   </ul>
```
jromero, line 217:
```suggestion
    <li><sup>&dagger;</sup> This feature does not apply to solution.</li>
    <li><sup>&ddagger;</sup> Requires additional configuration which adversely affects other features.</li>
    <li><sup>&diams;</sup> Requires additional configuration.</li>
  </ul>
```

Stylistically, the daggers should stay together. If you'd like to reorder the sentences then I would suggest that the symbols get changed to match.

jromero (changes requested):
Overall this is a great addition!

The columns are getting a little tight. In the future, I wonder if it would be better to have a sub-section just comparing it to Docker (and Dockerfile specific alternatives).

samj1912:
image

jromero (approved):

jromero:
docker

PR #385, opened 2021-08-03
Clarify buildpack description even more

abitrolly:
Signed-off-by: Anatoli Babenia <anatoli@rainforce.org>

I am trying to improve that once again.

### DIFF
--- a/content/docs/concepts/components/buildpack.md
+++ b/content/docs/concepts/components/buildpack.md
@@ -28,7 +28,7 @@ There are two essential steps that allow buildpacks to create a runnable image.
 
 #### Detect
 
-To determine which buildpack should be used to build your app, a [`platform`](platform) sequentially tests a group of
+To determine which buildpack should be used to build your app, a [`platform`][platform] sequentially tests a group of
 buildpacks against your source. The first buildpack that claims that it can build your sources will become selected.
 The `platform` does this by running `bin/detect` with a path to your source code, and expects the script to return 0
 if its logic matches -- for instance, an **NPM buildpack** might look for a `package.json`, and a **Go buildpack**
@@ -50,3 +50,4 @@ Learn more about buildpacks by referring to the [Buildpack API][buildpack-api].
 
 [buildpack-api]: /docs/reference/buildpack-api
 [package-a-buildpack]: /docs/buildpack-author-guide/package-a-buildpack/
+[platform]: /docs/concepts/components/platform
--- a/content/docs/concepts/components/buildpack.md
+++ b/content/docs/concepts/components/buildpack.md
@@ -5,16 +5,16 @@ weight=2
 
 ## What is a buildpack?
 
-A buildpack is a unit of work that inspects your app source code and formulates a plan to build and run your
+A buildpack is a set of executables that inspects your app source code and create a plan to build and run your
 application.
 
 <!--more-->
 
-Typical buildpacks are a set of at least three files:
+Typical buildpacks consist of at least three files:
 
 * `buildpack.toml` -- provides metadata about your buildpack
 * `bin/detect` -- determines whether buildpack should be applied
-* `bin/build` -- executes buildpack logic
+* `bin/build` -- executes build logic
 
 #### Meta-buildpack
 
@@ -22,16 +22,17 @@ There is a different type of buildpack commonly referred to as a **meta-buildpac
 `buildpack.toml` file with an `order` configuration that references other buildpacks. This is useful for 
 composing more complex detection strategies.
 
-## Anatomy of a buildpack
+## Buildpack phases
 
-There are two essential phases that allow buildpacks to create a runnable image.
+There are two essential steps that allow buildpacks to create a runnable image.
 
 #### Detect
 
-A platform sequentially tests groups of buildpacks against your app's source code. The first group that deems itself 
-fit for your source code will become the selected set of buildpacks for your app. Detection criteria is specific to each 
-buildpack -- for instance, an **NPM buildpack** might look for a `package.json`, and a **Go buildpack** might look for 
-Go source files.
+To determine which buildpack should be used to build your app, a [`platform`](platform) sequentially tests a group of
+buildpacks against your source. The first buildpack that claims that it can build your sources will become selected.
+The `platform` does this by running `bin/detect` with a path to your source code, and expects the script to return 0
+if its logic matches -- for instance, an **NPM buildpack** might look for a `package.json`, and a **Go buildpack**
+might look for Go source files.
 
 #### Build
 


content/docs/concepts/components/buildpack.md
```diff
@@ -5,33 +5,34 @@ weight=2
...
+These are two essential steps that allow buildpacks to create a runnable image.
 
 #### Detect
 
-A platform sequentially tests groups of buildpacks against your app's source code. The first group that deems itself 
-fit for your source code will become the selected set of buildpacks for your app. Detection criteria is specific to each 
-buildpack -- for instance, an **NPM buildpack** might look for a `package.json`, and a **Go buildpack** might look for 
-Go source files.
+To determine which buildpack should be used to build your app, a [`platform`](platform) sequentially tests a group of
+buildpacks against your source. The first buildpack that claims that it can build your sources will become selected.
```
samj1912, line 32: This may not always be true. The first buildpack group is selected but there could possibly be multiple buildpacks in that group that can build your source code.

content/docs/concepts/components/buildpack.md
```diff
@@ -5,33 +5,34 @@ weight=2
...
 
 There is a different type of buildpack commonly referred to as a **meta-buildpack**. It contains only a 
 `buildpack.toml` file with an `order` configuration that references other buildpacks. This is useful for 
 composing more complex detection strategies.
 
-## Anatomy of a buildpack
+## Buildpack phases
 
-There are two essential phases that allow buildpacks to create a runnable image.
+These are two essential steps that allow buildpacks to create a runnable image.
```
samj1912, line 27:
```suggestion
There are two essential steps that allow buildpacks to create a runnable image.
```

content/docs/concepts/components/buildpack.md
```diff
@@ -5,33 +5,34 @@ weight=2
 
 ## What is a buildpack?
 
-A buildpack is a unit of work that inspects your app source code and formulates a plan to build and run your
+A buildpack is a set of scripts that inspects your app source code and creates a plan to build and run your
```
samj1912, line 8:
```suggestion
A buildpack is a set of executables that inspects your app source code and create a plan to build and run your
```

content/docs/concepts/components/buildpack.md
```diff
@@ -5,33 +5,34 @@ weight=2
...
 
 #### Detect
 
-A platform sequentially tests groups of buildpacks against your app's source code. The first group that deems itself 
-fit for your source code will become the selected set of buildpacks for your app. Detection criteria is specific to each 
-buildpack -- for instance, an **NPM buildpack** might look for a `package.json`, and a **Go buildpack** might look for 
-Go source files.
+To determine which buildpack should be used to build your app, a [`platform`](platform) sequentially tests a group of
+buildpacks against your source. The first buildpack that claims that it can build your sources will become selected.
+The `platform` does this by running `bin/detect` with a path to your source code, and expects the script to return 0
```
samj1912, line 33: I think this line might be confusing in the context of buildpack groups and matching up of provisions and requirements within the group.

samj1912 (commented):
Thank you for the PR! :) I left some comments, lmk what you think about them.

content/docs/concepts/components/buildpack.md
abitrolly: I intentionally wrote this way as I thought that it would not repeat that is already stated in the header, but I am not proficient at the language. ) Fixed.

samj1912:
@abitrolly I was wondering if you have also had a chance to look at https://buildpacks.io/docs/concepts/#how-do-they-work

We tried to cut some details for a simpler and higher level overview of the different concepts. Does that flow better?

content/docs/concepts/components/buildpack.md
abitrolly: On Heroku I only used one buildpack for one build. This group logic is something new. It deserves a separate paragraph to keep it simple. Maybe it is documented somewhere already? How buildpacks share results of the build? Who creates the final image and packs artifacts then?

samj1912: They all work together to create the output image. See https://buildpacks.io/docs/reference/spec/buildpack-api/#build-plan

content/docs/concepts/components/buildpack.md
abitrolly: Buildpack groups is more or less clear. Maybe it needs a separate page though.  But the matching up of provisions and requirements within the group is not. Can you explain what is it?

content/docs/concepts/components/buildpack.md
samj1912: I am not too sure about how heroku uses CNB but I think they also allow multiple buildpacks https://devcenter.heroku.com/articles/using-multiple-buildpacks-for-an-app

content/docs/concepts/components/buildpack.md
samj1912: Some details are available at https://buildpacks.io/docs/reference/spec/buildpack-api/#build-plan

but essentially it requires all the buildpacks in the group to require everything that is provided by other buildpacks before them. Otherwise the group fails at detection.

abitrolly:
Haven't seen https://buildpacks.io/docs/concepts/#how-do-they-work before. It repeats much from the definition, and it doesn't explain the *buildpack groups* concept too. While it gives an abstract overview of what a buildpack does, it still doesn't explain what a buildpack is.

content/docs/concepts/components/buildpack.md
abitrolly: @samj1912 it looks like there is no obligation for the `buildpack` to provide build plan. And even if it provides, it just to pass info from own `detect` to own `build`, and further orchestration of buildpacks is the role of `lifecycle` tool, so I think it is okay to omit this here.

content/docs/concepts/components/buildpack.md
abitrolly: As for me, this definitely needs a separate page describing the scenario of multiple buildpacks.

abitrolly:
@samj1912 so do we reach a consensus that orchestration of multiple buildpacks is out of scope for buildpack definition page?

content/docs/concepts/components/buildpack.md
jromero: I agree with @samj1912, this sentence may be very misleading:

> The first buildpack that claims that it can build your sources will become selected.

It's really more like the "first group of buildpacks that claim they can..."

abitrolly: @jromero I am lost. How can a "group of buildpacks" collectively determine if they fit? Is it still `platform` that selects multiple buildpacks based on their individual responses? Or what is the "consensus algorithm" that makes buildpacks determine that they can only do this build together?

It is possible to document the simple case and then adjust for a more complicated scenario, which I believe needs a separate page.

jromero: Sorry for the delay. You can learn more about how buildpacks groups work here: https://buildpacks.io/docs/concepts/components/buildpack-group/

I agree with the sentiment that we could document a simple case and add more complicated scenarios later but the sentence in question is just wrong based on the existing functionality. 

A simple change from:

> The first buildpack that claims that it can build your sources will become selected.

To:

> The first group of buildpacks that claim to be able to build your sources will be used.

Should resolve the apprehension.

content/docs/concepts/components/buildpack.md
jromero: In addition to @samj1912 comments, "expects the script to return 0" seems to low-level of a detail for this page.

content/docs/concepts/components/buildpack.md
jromero: Looking back at was originally there, I think that makes good enough sense. If the reader needs more clarification, then maybe we can refer them to https://buildpacks.io/docs/concepts/components/buildpack-group/.

abitrolly:
Now it is much better. Thanks everyone involved. )

jromero:
No. _Thank you_!

PR #390, opened 2021-08-17
Add conventions for app source metadata

jromero:
Based on discussion: https://github.com/buildpacks/community/discussions/120#discussioncomment-1082492

localhost_1313_docs_platform-guide_conventions_source-metadata_


samj1912 (approved):
lgtm. Did we decide to not go with the array version?

jromero:
> lgtm. Did we decide to not go with the array version?

If I understand what you are talking about, that would require and RFC and spec update. I'm still in favor of that but need to move forward with what we have now.

PR #407, opened 2021-09-05
feat: install latest hugo according to platform

Harshita-Kanal:
### Summary
The approach here uses the GitHub API to fetch the `browser_download_url` which is obtained by parsing the JSON response. For windows, this gets the latest zip file of Hugo extended for windows OS. However, the implementation for Mac, Linux is a work in progress.

Signed-off-by: Harshita-Kanal <harshita.kgv@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,17 @@ clean:
 .PHONY: install-hugo
 install-hugo:
 	@echo "> Installing hugo..."
-	cd tools; go install -mod=mod --tags extended github.com/gohugoio/hugo
+	@echo "This may take a while depending on your connection."
+ifeq ($(OS),Windows_NT)
+	@echo $(shell uname -s)	
+	cd tools; mkdir bin; cd bin; curl -L -O $(shell curl https://api.github.com/repos/gohugoio/hugo/releases/latest | jq -r '.assets[30].browser_download_url')
+	
+else
+	@echo $(shell uname -s)
+	cd tools; mkdir bin; cd bin; curl -L -O $(shell curl https://api.github.com/repos/gohugoio/hugo/releases/latest | jq -r '.assets[26].browser_download_url')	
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -29,7 +29,17 @@ clean:
...
 install-hugo:
 	@echo "> Installing hugo..."
-	cd tools; go install -mod=mod --tags extended github.com/gohugoio/hugo
+	@echo "This may take a while depending on your connection."
+ifeq ($(OS),Windows_NT)
+	@echo $(shell uname -s)	
+	cd tools; mkdir bin; cd bin; curl -L -O $(shell curl https://api.github.com/repos/gohugoio/hugo/releases/latest | jq -r '.assets[30].browser_download_url')
+	
+else
+	@echo $(shell uname -s)
```
jromero, line 38: Taken from https://stackoverflow.com/a/12099167/552902

Something like this should work.

```
UNAME_S := $(shell uname -s)
ifeq ($(UNAME_S),Darwin)
   // macOS
else
   // assumed linux
endif
```

jromero (changes requested):
Note: We'll also want to add something like:

```Makefile
HUGO_BIN=./tools/bin/hugo
```

and then use that variable when calling the `hugo` executable anywhere in the Makefile.

jromero:
@Harshita-Kanal 👈 (poke)

Is this something you are still interested in resolving. It looks to be so close to being done.

Harshita-Kanal:
> @Harshita-Kanal 👈 (poke)
> 
> Is this something you are still interested in resolving. It looks to be so close to being done.

Yes just missed it, sorry for the delay!

Harshita-Kanal:
Was able to get it to work for Linux I think. How should we handle the case for Windows?

PR #413, opened 2021-09-11
Add Features top-level section in /docs

Harshita-Kanal:
## Summary 
The PR adds a top level features section and moves documentation for `reproducible builds` there.

Signed-off-by: Harshita-Kanal <harshita.kgv@gmail.com>


jromero (changes requested):
🙏 Thank you!

If I'm not mistaken, the reproducibility file was copied from [here](https://github.com/buildpacks/docs/blob/main/content/docs/reference/reproducibility.md). 

Let's remove the old location so that we don't have duplicate content. 

🍪 Bonus points if you put an alias pointing to where it used to be (we have a few examples if you search for [aliases](https://github.com/buildpacks/docs/search?q=aliases)).

Harshita-Kanal:
Thanks! made the changes :)

jromero (approved):

PR #417, opened 2021-09-21
Document how pack interacts with http proxies

AidanDelaney:
Show how to both build an application behind a proxy and how an
application can be configured to access web APIs through a proxy.

### DIFF
--- a/content/docs/app-developer-guide/using-http-proxy.md
+++ b/content/docs/app-developer-guide/using-http-proxy.md
@@ -21,11 +21,16 @@ ERROR: failed to build: failed to fetch builder image 'index.docker.io/cnbs/samp
 : Error response from daemon: Get "https//registry-1.docker.io/v2/": context deadline exceeded
 ```
 
-The `pack` tool uses the Docker daemon to manage the local image registry on your machine.  The `pack` tool will ask the Docker daemon to download buildpacks and images for you.  Because of this relationship, between `pack` and the Docker daemon, we need to configure the Docker daemon to use a HTTP proxy.
+The `pack` tool uses the Docker daemon to manage the local image registry on your machine.  The `pack` tool will ask the Docker daemon to download buildpacks and images for you.  Because of this relationship, between `pack` and the Docker daemon, we need to configure the Docker daemon to use a HTTP proxy.  The approach to setting the HTTP proxy depends on your platform:
 
-The Docker project documents [how to configure configure the HTTP/HTTPS proxy](https://docs.docker.com/config/daemon/systemd/#httphttps-proxy) settings for the Docker daemon.  You should configure the `HTTP_PROXY` and `HTTPS_PROXY` environment variables as part of the Docker daemon startup.
 
-### Proxy Settings for Buildpacks
+### Docker Desktop (Windows and MacOS)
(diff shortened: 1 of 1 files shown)


AidanDelaney:
Some buildpack operations [require the `http_` env vars](https://github.com/buildpacks/pack/blob/437c3f15a7501d432be94ebeabfb119a396ca549/build.go#L599) to be set.  This is independent of setting `~/.docker/config.json`.

I had expected that it is sufficient to set `httpProxy` in `~/.docker/config.json`.  However, when running `pack build --verbose sample-app --path samples/apps/java-maven --builder cnbs/sample-builder:bionic` the buildpack fails to install the JDK unless the `http_*` vars are also set.

content/docs/app-developer-guide/using-http-proxy.md
```diff
@@ -0,0 +1,54 @@
...
+
+```console
+$ pack build sample-app --path samples/apps/java-maven --builder cnbs/sample-builder:bionic
+ERROR: failed to build: failed to fetch builder image 'index.docker.io/cnbs/sample-builder:bionic'
+: Error response from daemon: Get "https//registry-1.docker.io/v2/": context deadline exceeded
+```
+
+The `pack` tool uses the Docker daemon to manage the local image registry on your machine.  The `pack` tool will ask the Docker daemon to download buildpacks and images for you.  Because of this relationship, between `pack` and the Docker daemon, we need to configure the Docker daemon to use a HTTP proxy.
+
+The Docker project documents [how to configure configure the HTTP/HTTPS proxy](https://docs.docker.com/config/daemon/systemd/#httphttps-proxy) settings for the Docker daemon.  You should configure the `HTTP_PROXY` and `HTTPS_PROXY` environment variables as part of the Docker daemon startup.
```
jromero, line 26: The link is linux specific which I find confusing as a Docker Desktop user. Based on [this link](https://docs.docker.com/desktop/mac/#resources), there are 2 ways to configure it for Docker Desktop. 

1. Setting [a proxy at the OS level](https://support.apple.com/guide/mac-help/enter-proxy-server-settings-on-mac-mchlp2591/mac) would be also propagate to the daemon on on Docker Desktop.
    > Docker Desktop detects HTTP/HTTPS Proxy Settings from macOS and automatically propagates these to Docker. 
2. Setting the Proxy settings in the [Resources](https://docs.docker.com/desktop/mac/#resources) section of Preferences.

Maybe it's worth denoting the difference based on the user setup. Docker Desktop and Linux.

jromero (commented):
I really like the overall structure and information presented. Left a slight comment on what I think could be clarified. Let me know what you think.

content/docs/app-developer-guide/using-http-proxy.md
AidanDelaney: Thank you.  Yes, I will differentiate between Linux and Docker Desktop.

AidanDelaney: In "Making `pack` proxy aware" I've provided Docker Desktop and Linux instructions.  I thought that was a better distinction than between Docker Desktop and Docker CE as some developers may be unfamiliar with the distinction between Desktop and CE.  I used a link to `dummies.com` for the Windows proxy configuration as I could not find an end-user focused guide in the official MS support documentation.

samj1912 (approved):

jromero (approved):

PR #418, opened 2021-09-21
Call out the default project descirptor location.

mboldt:
Without that, the example is confusing since it does not use the `--descriptor` flag.

Signed-off-by: Mikey Boldt <mboldt@vmware.com>


jromero (approved):

jromero:
🙏  Thanks for the contribution!

PR #420, opened 2021-09-26
Changed run image text to life cycle 

### LINKED ISSUE #415: Correct builder documentation
## Summary

The [concepts page](https://buildpacks.io/docs/concepts/#what-is-a-builder) on the docs says this:
> Builders are an ordered combination of buildpacks with a base build image and a **run image**.

While the [builders page](https://buildpacks.io/docs/concepts/components/builder/) has this:
> A builder is an image that contains all the components necessary to execute a build. A builder image is created by taking a build image and adding a lifecycle, buildpacks, and files that configure aspects of the build including the buildpack detection order and the location(s) of the run image

So the *run image* should be replace with "lifecycle", unless I'm misunderstanding the concept.

## Additional Information

Can be labeled as good first issue

SanjaySinghRajpoot:
Fixes #415


content/docs/concepts/components/builder.md
```diff
@@ -1,7 +1,7 @@
 +++
 title="Builder"
 weight=1
-summary="A builder is an image that contains all the components necessary to execute a build. A builder image is created by taking a build image and adding a lifecycle, buildpacks, and files that configure aspects of the build including the buildpack detection order and the location(s) of the run image"
+summary="A builder is an image that contains all the components necessary to execute a build. A builder image is created by taking a build image and adding a lifecycle, buildpacks, and files that configure aspects of the build including the buildpack detection order and the location(s) of the lifecycle"
```
samj1912, line 4: This doesn't need to be changed. This was previously correct.

SanjaySinghRajpoot: ok.

PR #429, opened 2021-11-09
Add packaging con 2021 slides

samj1912:


dfreilich (approved):
Looking forward to the recording being available! I had shared these slides with my teammates, and they were super impressed with how clear and thorough these are, so this is really a great resource :D

jromero (approved):

PR #431, opened 2021-11-17
Implement redesign of feature comparison

Harshita-Kanal:
## Summary
Adds initial layout for 1 on 1 comparison. (Work in progress)


Harshita-Kanal:
> @Harshita-Kanal, this is looking great! 😍
> 
> Are you opposed to me creating an "epic" branch where we can start merging all the PRs into? This would allow us to focus on small chunks of work at a time.

Yeah sure, that would be great 😄

PR #433, opened 2021-11-22
Add -stack to removed flags in platform api 0.7

samj1912:


samj1912:
Actually, a bit confused now. https://github.com/buildpacks/spec/blob/platform/0.7/platform.md#exporter doesn't mention stack as a possible flag to exporter but looking at the lifecycle code and platform 0.8, it does exist.

samj1912:
Ah, git blame to the rescue https://github.com/buildpacks/spec/pull/250

samj1912:
I would prefer if we went back and fixed older released specs with typos.

natalieparellano:
This should be closed, no? Since the spec is wrong...

PR #462, opened 2022-02-07
Fix typo in name of flag

nokome:
Small typo fix.

PR #464, opened 2022-02-10
Fix typo in name of flag

AidanDelaney:
Fix typo in name of flag and fold typo fix in to katacoda
docs.


AidanDelaney:
Pulls in changes from #462 and ensures that the build passes.

samj1912 (approved):

PR #467, opened 2022-02-18
Fix buildpack caching logic for API 0.6+

samj1912:


content/docs/buildpack-author-guide/create-buildpack/adding-bill-of-materials.md
```diff
@@ -161,10 +163,6 @@ else
     echo "---> Installing gems"
     mkdir -p "$bundlerlayer"
     cat > "$layersdir/bundler.toml" << EOL
```
leshik, line 165: Here it should append IMO, otherwise, it will overwrite the previously set `[types]` table.

leshik (changes requested):

AidanDelaney (approved):
This is great.  Thanks.

PR #478, opened 2022-03-22
Update docs for Buildpack API 0.7 -> 0.8

mboldt:


content/docs/reference/spec/buildpack-api.md
```diff
@@ -52,7 +53,7 @@ if [ -f requirements.txt ]; then
   echo "Python Buildpack"
   exit 0
 else
-  exit 1
+  exit 100
```
AidanDelaney, line 56: Do we use exit code 100?

content/docs/reference/spec/migration/buildpack-api-0.7-0.8.md
```diff
@@ -0,0 +1,52 @@
...
+
+This guide is most relevant to buildpack authors.
+
+See the [spec release](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.8) for buildpack API 0.8 for the full list of changes and further details.
+
+### Process-Specific Working Directory
+
+Buildpacks may specify the working directory for a process by setting the `working-dir` field on the process in [`launch.toml`](https://github.com/buildpacks/spec/blob/buildpack/0.8/buildpack.md#launchtoml-toml).
+
+Prior to this, all processes ran from the app directory (`CNB_APP_DIR`).
```
AidanDelaney, line 15: Make it clear that `CNB_APP_DIR` is a build time variable:
```suggestion
Prior to this, all processes ran from the app directory (configurable at image build time through `CNB_APP_DIR`).
```

content/docs/reference/spec/migration/buildpack-api-0.7-0.8.md
```diff
@@ -0,0 +1,52 @@
...
+command = "<start-process>"
+working-dir = "<working-dir>"
+```
+
+For details, see [RFC 81](https://github.com/buildpacks/rfcs/blob/main/text/0081-process-specific-working-directory.md).
+
+### Deprecate Positional Args to `build` and `detect` Executables
+
+The positional arguments to the `detect` and `build` executables are deprecated.
+The lifecycle now provides these values as environment variables.
```
AidanDelaney, line 37:
```suggestion
Lifecycle now accepts these values as environment variables.
```

AidanDelaney (changes requested):
Thanks, this is great.  I have one question and two suggestions that you do not have to accept.  After that, it's LGTM.

content/docs/reference/spec/buildpack-api.md
mboldt: [Per spec](https://github.com/buildpacks/spec/blob/main/buildpack.md#detection), 100 is failed detection (as in, this buildpack does not apply to this app) and other non-zero indicates an error during detection.

content/docs/reference/spec/migration/buildpack-api-0.7-0.8.md
mboldt: As I understand it, `CNB_APP_DIR` must be set to the app directory by the platform, as input to the lifecycle phases (builder, detector, exporter, launcher). I don't think it's configurable from the buildpack author / buildpack API perspective.

That said, perhaps `CNB_APP_DIR` here is not clarifying, and we could just omit the reference to it. What do you think?

content/docs/reference/spec/buildpack-api.md
samj1912: Ah, didn't catch that. Yup, should be okay to switch to 100 then.

AidanDelaney (approved):

jromero (approved):
LGTM but I'd defer to @samj1912 to merge due to the knowledge required to review the details.

samj1912:
Marking it as draft meanwhile so we don't accidentally merge this. The contents look good, just not ready for a merge.

PR #479, opened 2022-03-22
Updated pack version in katacoda

AidanDelaney:
Update the version of `pack` that is used in
our katacoda guide.


samj1912 (changes requested):
We should also update this for the other katacoda scenarios.

AidanDelaney:
> We should also update this for the other katacoda scenarios.

Updated other `background.sh` files.

samj1912 (approved):

PR #506, opened 2022-07-07
Note that image builders do not exist for arm64

josegonzalez:
The previous version said no arm64 platforms do not exist, which is no longer true. `pack` now supports arm64.

Signed-off-by: Jose Diaz-Gonzalez <email@josediazgonzalez.com>


AidanDelaney (approved):
Thanks @josegonzalez for this!

jromero (approved):

PR #513, opened 2022-08-08
Address issues in caching description

### LINKED ISSUE #512: Incorrect documentation on caching strategies
**On Page:** [Caching Strategies](https://buildpacks.io/docs/buildpack-author-guide/caching-strategies/)

On the page we note a couple of things that could be misleading or incorrect.

We could also probably title the page as layer types instead of caching strategies. 

> ignored layers are available to buildpack authors for use as temporary layers.

We should either remove references to the "ignored" layer type or note that an ignored layer is any layer with all of the other 3 types set to false. It's not a layer boolean that exists in the spec and might confuse users. 

> However, we want to invalidate previously cached layers if the buildpacks API changes,
the type of the layer changes.A layer built using a buildpack at API version 0.7 should be considered invalid if the API version for that buildpack has been updated to 0.8. Similarly, if a layer is changed from being a cache-only layer to being a cache and launch layer, then the cache should be considered invalid.

The statements about layers being invalidated due to buildpack api changes or layer type changes are incorrect and contradictory to what existing production buildpacks do. This is in fact something we actively think about preserving during buildpacks api changes and implementation in the lifecycle. It is completely okay and valid for buildpacks to reuse layers between different buildpack APIs. We also allow layers to be reused when layer types are changed. A cache=true, launch=true layer will be restored from cache even if we add a build flag to it. 

> In addition to general cache invalidation conditions a buildpack should invalidate a previous layer if an important property changes, such as:

> the major version of the runtime changes eg: NodeJS changes from 16 to 18
requested application dependencies have changed eg: a Python application adds a dependency on the requests module 

Any examples that currently use the word "should" could be changed to "could". The page demonstrates possible examples and should make it clear that they are just examples and not recommendations. 

The jq example being used is also incorrect to demonstrate proper usage of the cache flag.

> Setting cache = true allows additional fine-grained control over caching. The cache = true flag caches a layer and allows a buildpack to make content level decisions about the validity of the cache (as opposed to using the less granular metadata). As an example, suppose a layer where launch = true installs a jq binary with version 1.5 and sets version=1.5 in the layer metadata. By default, this layer will not be re-used from the registry when a buildpack requests jq with version=1.6 to be installed. However, setting cache = true makes a previously built layer available during the build. A buildpack could then prefer to implement logic to restore jq with version=1.5 instead of performing a potentially expensive download of jq with version=1.6. The cache = true setting allows for cache validation decisions to be made at a level of granularity that is much finer grained than layer metadata.

The above can also be achieved by using a launch only layer. Launch only layer restoration is not dependent on metadata changes. A launch layer is marked for reuse by re-setting launch=true during the next build and by leaving the layer contents empty. Changing the layer metadata does not affect layer reuse.

> Launch layers are exported to an OCI registry and layer metadata is stored with the launch layer.

This statement can be a bit misleading. Layer metadata for launch layers is not stored with launch layers. Rather it's stored in a separate layer in the app image.  This is however an implementation detail and I think we should just mention that layer metadata can be stored in the output image. It can however be stored/restored either from the cache or the output image based on the later types.

AidanDelaney:
Fixed #512

### DIFF
--- a/content/docs/buildpack-author-guide/caching-strategies.md
+++ b/content/docs/buildpack-author-guide/caching-strategies.md
@@ -35,7 +35,7 @@ This will guarantee that the previously published application image layer in the
 
 Setting `build = true` makes a layer available to subsequent buildpacks.  Therefore binaries installed to the `bin` directory on a `build = true` layer are available to subsequent buildpacks during the build phase.  It is also the case that `lib` directories on a `build = true` later are added to the `LD_LIBRARY_PATH` during the build phase of subsequent buildpacks.  Environment variables defined in a `build = true` layer are similarly available.  For any layer where `launch = true` and `build = true`, a launch layer from the OCI registry can no longer be reused. Instead, the layer must be made available locally so that subsequent buildpacks can use it.
 
-Setting `cache = true` keeps a local cache of the layer.  The cache is restored before the buildpacks build phase.  
+Setting `cache = true` ensures that the layer is restored locally before the buildpacks build phase.
 
 Setting `cache = false`, `build = false`, and `launch = true` is the most common configuration.  If `cache = false`, `build = false`, and `launch = true` is not appropriate for your layer, then `cache = true`, `build = true`, and `launch = true` should be the next combination to evaluate:
 
--- a/content/docs/buildpack-author-guide/caching-strategies.md
+++ b/content/docs/buildpack-author-guide/caching-strategies.md
@@ -1,10 +1,10 @@
 +++
-title="Caching Strategies"
+title="Layer Types"
 weight=6
 summary="Learn strategies for caching layers."
 +++
 
-# Caching
+# Layers
 
 There are three types of layers that can be contributed to an image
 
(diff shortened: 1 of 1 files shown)


content/docs/buildpack-author-guide/caching-strategies.md
```diff
@@ -48,7 +35,7 @@ This will guarantee that the previously published application image layer in the
...
 
-Setting `cache = true` allows additional fine-grained control over caching.  The `cache = true` flag caches a layer and allows a buildpack to make _content_ level decisions about the validity of the cache (as opposed to using the less granular metadata).  As an example, suppose a layer where `launch = true` installs a `jq` binary with version `1.5` and sets `version=1.5` in the layer metadata.  By default, this layer will not be re-used from the registry when a buildpack requests `jq` with `version=1.6` to be installed.  However, setting `cache = true` makes a previously built layer available during the build.  A buildpack could then prefer to implement logic to restore `jq` with `version=1.5` instead of performing a potentially expensive download of `jq` with `version=1.6`.  The `cache = true` setting allows for cache validation decisions to be made at a level of granularity that is much finer grained than layer metadata.
+Setting `cache = true` keeps a local cache of the layer.  The cache is restored before the buildpacks build phase.
```
samj1912, line 38: The cache may or may not be local. Cached layers can also be stored in the registry as a cache image.

samj1912: Setting `cache = true` implies that the layer is restored locally before the buildpacks build phase.

samj1912 (approved):

jkutner (approved):

PR #542, opened 2022-12-15
Fix upgrade pack in CI

natalieparellano:
While we await resolution of https://github.com/buildpacks/pack/issues/1574 this should fix the pack upgrade

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,10 +24,7 @@ jobs:
       - name: Install Dependencies
         run: sudo apt-get install make curl jq
       - name: Install pack
-        run: |
-          sudo add-apt-repository ppa:cncf-buildpacks/pack-cli
-          sudo apt-get update
-          sudo apt-get install pack-cli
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -23,6 +23,11 @@ jobs:
           go-version: '1.x'
       - name: Install Dependencies
         run: sudo apt-get install make curl jq
+      - name: Install pack
```
samj1912, line 26: Should we instead use https://github.com/buildpacks/github-actions#setup-pack-cli-action ?

.github/workflows/main.yml
```diff
@@ -23,6 +23,11 @@ jobs:
...
+        run: |
+          sudo add-apt-repository ppa:cncf-buildpacks/pack-cli
+          sudo apt-get update
+          sudo apt-get install pack-cli
```
samj1912, line 30:
```suggestion
        uses: buildpacks/github-actions/setup-pack@v4.1.0
```

samj1912 (approved):

PR #543, opened 2022-12-22
Revert "Promote guides to second level header"

samj1912:
Reverts buildpacks/docs#539

I just noticed that this section bleeds into the homepage and looks very weird with multiple top navigation bars.

Screenshot_2022-12-22-18-50-39-38_40deb401b9ffe8e1df2f1cc5ba480b12

Screenshot_2022-12-22-18-51-50-06_40deb401b9ffe8e1df2f1cc5ba480b12

It also isn't responsive and looks bad on mobile screens. Reverting it for now until we can figure out a better way to display this. Ideally we should have something like - 
Screenshot_2022-12-22-18-45-13-05_abb9c8060a0a12c5ac89e934e52a2f4f

Notice the 'which are you' section 

https://web.archive.org/web/20210511112502/https://www.seldon.io/tech/

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -81,21 +81,6 @@ featureKatacoda = false
   url        = "https://github.com/buildpacks"
   weight     = 4
 
-  [[menu.docs]]
-  name       = "Application Developer Guide"
-  url        = "/docs/app-developer-guide"
-  weight     = 1
-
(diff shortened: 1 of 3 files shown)


AidanDelaney (approved):
Let's revert for the moment

PR #546, opened 2022-12-27
docs: correct typo and expected Ruby version

LayZeeDK:
Signed-off-by: Lars Gyrup Brink Nielsen <larsbrinknielsen@gmail.com>


AidanDelaney:
Thanks for the PR @LayZeeDK.  The build is failing because our docs build system is a little complex to support a feature we no-longer use.  I'll remove the dependency on that older feature and then pull in your change.

AidanDelaney:
Will build if we land https://github.com/buildpacks/docs/pull/548

samj1912 (approved):

LayZeeDK:
Thank you, @AidanDelaney and @samj1912 🙇‍♂️

PR #565, opened 2023-02-10
Move inline buildpacks docs to Extension Author Guide

AidanDelaney:
Take advantage of the new Dockerfiles extensions docs to consolidate our docs around extending the build process in a single section.

Signed-off-by: Aidan Delaney <adelaney21@bloomberg.net>

### DIFF
--- a/content/docs/extension-author-guide/_index.md
+++ b/content/docs/extension-author-guide/_index.md
@@ -1,5 +1,5 @@
 +++
-title="Extension Author Guide"
+title="Extending Buildpacks"
 weight=4
 include_summaries=true
 expand=true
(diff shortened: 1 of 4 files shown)


AidanDelaney:
@natalieparellano This builds on your guide, so I'd appreciate your opinion?

natalieparellano:
I'm thinking back to the "4 quadrants of docs" and while thematically I think it makes sense to group inline buildpacks and extensions together if your "how to" question is "how do I configure my buildpacks build" in terms of reference content inline buildpacks and extensions actually have very little to do with each other. Maybe we should make a whole separate "landing page" for "buildpacks extension points" and point to each of inline buildpacks and authoring an extension as sub bullets to that theme. In terms of persons I think where inline buildpacks are currently referenced (app developer guide) makes sense.

AidanDelaney:
Let's close this and take a different approach.

PR #571, opened 2023-03-03
Improve extensions demo 

natalieparellano:
...by not requiring the builder to be recreated after every failed build

The hello-extensions buildpack can be configured through the environment after https://github.com/buildpacks/samples/pull/142 lands


natalieparellano:
Folding this into https://github.com/buildpacks/docs/pull/581

PR #573, opened 2023-03-06
Fix hugo download on macOS

### LINKED ISSUE #572: Bug - when hugo is not installed on tools/bin folder and you are running macos make build fails
## Summary

Running `make build` on macOS (Darwring) when hugo is not installed

```bash
➜  > git:(main) ✗ make build
> Installing hugo for macOS (universal)...
curl -sSL -o tools/bin/hugo.tar.gz null
curl: (6) Could not resolve host: null
make: *** [tools/bin/hugo] Error 6
```

## Additional Information

If you want to reproduce this error, you just need to remove your hugo folder from `tools/bin/hugo` and try to build. For what I checked on the `Makefile` we are parsing the installer from this [url](https://api.github.com/repos/gohugoio/hugo/releases/latest) then we try to create the correct file name for macOS, checking the json file I noticed the filename has the following format:

```json
"name": "hugo_extended_0.111.2_darwin-universal.tar.gz",
```
Maybe, in the past the name of the OS was `macOS` but now it seems to be `darwing` that's why the url is not well formed

jjbustamante:
This PR fixes an error downloading hugo on macOS when no installation is available

Fixes #572

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ endif
 endif
 
 ifeq ($(shell uname -s),Darwin)
-	HUGO_OS:=macOS
+	HUGO_OS:=darwin
 	HUGO_ARCH:=universal
 endif
 


AidanDelaney (approved):
LGTM

PR #574, opened 2023-03-06
OCI layout documentation

### LINKED ISSUE #529: Document Export to OCI layout feature
## Summary

We need to document guide with **how to** use the new export to OCI layout feature

- As a developer, what steps do I need to do to export my application image on disk in OCI layout format?

- [x] App Developer
- [ ] Buildpack Author
- [ ] Operator
- [ ] Other:

## Additional Information

- Media Types: see this [comment](https://github.com/buildpacks/rfcs/pull/203#discussion_r1092449172)

jjbustamante:
The following pull requests adds documentation on how to use the new experimental feature to export the application image to disk in OCI layout format

I added a new section *Experimental* under the *Feature* section a it looks like this:

Screenshot 2023-08-17 at 8 23 33 AM

fixes #529


content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
...
+summary="Learn how to export your application image to disk in OCI layout format"
++++
+
+<div class="quote mb-4">
+    The OCI Image Layout is the directory structure for OCI content-addressable blobs and location-addressable references.
+    <div class="author">See the <a href="https://github.com/opencontainers/image-spec/blob/main/image-layout.md">specification</a></div>
+</div>
+
+Exporting to OCI layout format is an **experimental** feature available on pack since version X.Y.Z
+
```
AidanDelaney, line 13: I think we need a table somewhere comparing the `cnb` layout with the `oci` layout.  The first question I'd ask as a user is "Which layout should I choose?", and such a table would contain enough "Why?" to answer which layout to choose.

content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
...
+```bash
+pack config experimental true
+```
+
+You can confirm everything is fine, checking the `config.toml` file in your `PACK_HOME` installation folder, for example:
+
+```bash
+cat ~/.pack/config.toml
+experimental = true
+layout-repo-dir = "<$HOME>/.pack/layout-repo"
```
AidanDelaney, line 33: Does `pack config experimental true` add the "layout-repo-dir" to `config.toml`?

content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+The configuration shows the experimental mode was **enabled** and a local directory to save images on disk was configured to path `<$HOME>/.pack/layout-repo`.
+
+### 2. Build the app
+
+If you haven't already, please follow the steps to [build an app](/docs/app-developer-guide/build-an-app), once you managed to build a sample application you can try exporting the sample application to disk in OCI layout format.
```
AidanDelaney, line 40:
```suggestion
Please first follow the steps to [build an app](/docs/app-developer-guide/build-an-app), once you have successfully built an application you can export the sample application to disk in OCI layout format. 
```

We try and stay away from words like "managed", or terminology that suggests the process might be difficult.

content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
...
+
+The OCI layout feature must be enabled using the convention `oci:<path/to/save/image>` in the `<image-name>` parameter when invoking `pack build`.
+
+For example:
+
+```bash
+pack build oci:sample-app --path samples/apps/java-maven --builder cnbs/sample-builder:bionic
+```
+
+It will save the image in a folder `./sample-app` created in your current directory.
```
AidanDelaney, line 50: Why do we need `layout-repo-dir` if the OCI layout is saved to the PWD?

content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
...
+[runtime bundler](https://github.com/opencontainers/runtime-spec) that can be executed with another tool like [runc](https://github.com/opencontainers/runc)
+
+---
+
+## Extra configuration
+
+### Skip saving your run-image layers on disk
+
+Before using this option we suggest to remove your local layout directory (the one configured in your pack config.toml with the key `layout-repo-dir`) and 
+your application image folder (if you are planning to use the same folder), the reason for this is pack doesn't remove the blobs saved in the `layout-repo-dir` if you use the `--sparse` flag
```
AidanDelaney, line 90: Here it becomes more unclear why `layout-repo-dir` was necessary above.

content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
...
+│     ├── e1048fb89c3194a1f0542c0847aa086a7034dd7867c48fe8c93675cf36f90610
+│     ├── f0a30c5bc44742065b1b4ffa95271a39994f05ba7a03dd7e7143d1d3e45fa0b1
+│     └── f9d6350d0c44c0e7165a522155f53181ce8c163a6b8ead1f6baea22d1a8d8a78
+├── index.json
+└── oci-layout  
+
+3 directories, 13 files
+```
+If you are interested on keep playing with the image in  OCI layout format, one tool you can take a look is [umoci](https://umo.ci/) it can help you to create a 
+[runtime bundler](https://github.com/opencontainers/runtime-spec) that can be executed with another tool like [runc](https://github.com/opencontainers/runc)
```
AidanDelaney, line 81: There a number of tools here that we might need to explain to end-users.  I *think* we're using `umoci` because it is a strict-OCI compliant tool?  That assembles an image.  I _presume_ that docker can also execute a strict-OCI image?  If so, should we prefer to not introduce a new runtime there?

content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
...
+skopeo copy -a docker://<your-image> <dest>
+```
+It will give you `application/vnd.oci.image.index.v1+json` media type, which is currently working
+
+But [crane](https://github.com/google/go-containerregistry/tree/main/cmd/crane) 
+
+```bash
+crane pull <your-image> <dest> --format=oci
+```
+It will give you `application/vnd.docker.distribution.manifest.list.v2+json`, which will fail
```
AidanDelaney, line 146: Do these things pass/fail due to the state of _our_ current implementation?  Or because of a limitation in `crane`, or because the operation violates part of the OCI spec? Or for some other reason?

AidanDelaney (changes requested):
This is really great.  I wonder if we should push most of our docs about "experimental" features to a new section?  It is the case that only advanced users would be interested in these experimental features.

content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
++++
+title="Export to OCI layout format"
+weight=3
+summary="Learn how to export your application image to disk in OCI layout format"
++++
+
+<div class="quote mb-4">
+    The OCI Image Layout is the directory structure for OCI content-addressable blobs and location-addressable references.
+    <div class="author">See the <a href="https://github.com/opencontainers/image-spec/blob/main/image-layout.md">specification</a></div>
```
joe-kimmel-vmw, line 9:
```suggestion
    <div class="author">See the <a href="https://github.com/opencontainers/image-spec/blob/main/image-layout.md">specification.</a></div>
```

content/docs/app-developer-guide/export-to-oci-layout.md
joe-kimmel-vmw: agree, helpful to add a little more detail here

content/docs/app-developer-guide/export-to-oci-layout.md
jjbustamante: Yes, that's correct

content/docs/app-developer-guide/export-to-oci-layout.md
jjbustamante: We are using the `layout-repo-dir` to save the `run-image` on disk because actually invoking the lifecycle.  When lifecycle is configured to use OCI layout it is expecting to have all the require inputs (like the run-image) on disk,  because that should be transparent for users, we are saving those images in that particular folder.

content/docs/app-developer-guide/export-to-oci-layout.md
jjbustamante: Yeah, I had some doubts about including those tool, maybe we can suppose users know what they can do with the OCI layout format

content/docs/app-developer-guide/export-to-oci-layout.md
jjbustamante: It fails because of the state or our current implementation, it should be improve on new versions of the feature. I will add it

jjbustamante:
> This is really great. I wonder if we should push most of our docs about "experimental" features to a new section? It is the case that only advanced users would be interested in these experimental features.

@AidanDelaney 

Maybe I can move the documentation under *Feature* section, put something like *Experimental* there?

Screenshot 2023-05-05 at 3 20 54 PM

content/docs/app-developer-guide/export-to-oci-layout.md
jjbustamante: Let me explain this part a little better so you can guide me on how to document it. 

According to the [RFC](https://github.com/buildpacks/rfcs/blob/main/text/0119-export-to-oci.md#how-it-works) if you want to export your application image on disk, you need to mount to the lifecycle a *local storage* with the images require by the build process, for example, the `run-image`. This storage must contain images in OCI layout format and name resolution will follow some rules defined also in the RFC. 

The documented part was based on a feedback provided by @natalieparellano [here](https://github.com/buildpacks/rfcs/pull/203#discussion_r1092449172), which is basically the following scenario: Suppose you want to execute the lifecycle directly without using pack, for example, a platform implementor. They need to create this local storage and save everything there before executing the build, right?, well the problem is: *If they save in the local storage a manifest list* and then they execute the lifecycle, right now, we are not smart enough to detect which manifest we need to pick to execute the build. 

That is the actual comment we want to document so users can take it into consideration, now I am not sure this is the right place to do it, or maybe I didn't write it correctly, so any advice will be appreciate!

natalieparellano:
What's the status on this one? Does it still need changes?

content/docs/app-developer-guide/export-to-oci-layout.md
```diff
@@ -0,0 +1,151 @@
...
+<div class="quote mb-4">
+    The OCI Image Layout is the directory structure for OCI content-addressable blobs and location-addressable references.
+    <div class="author">See the <a href="https://github.com/opencontainers/image-spec/blob/main/image-layout.md">specification.</a></div>
+</div>
+
+Exporting to OCI layout format is an **experimental** feature available on pack since version X.Y.Z
+
+### 1. Enable experimental feature
+
+Verify your pack version is equal or greater than X.Y.Z
```
AidanDelaney, line 16:
```suggestion
Verify your pack version is equal or greater than 0.29.0
```

AidanDelaney (approved):
LGTM.  Thanks a lot.  You raise a great point about the structure of our docs.  I'll think about this.

PR #582, opened 2023-04-05
deprecate stacks, add targets

joe-kimmel-vmw:
a pass at revising our docs for targets.  hopefully provides a starting point to refine them into something we all want to publish. 

https://github.com/buildpacks/docs/issues/422


content/docs/concepts/components/stack.md
```diff
@@ -6,7 +6,48 @@ aliases=[
...
 +++
 
-## What is a stack?
+## What is a stack and Why is it deprecated?
+
+### Deprecation
+Stacks are great - long live the stack! 
+However, a stack is basically an extra layer on top of the **build image** (where your app is built)
+and the **run image** (where your app runs) and [the community decided](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md)
+to use those two concepts directly without the additional layer formerly known as _stack_.
```
AidanDelaney, line 15: I wonder if it makes sense to compare a stack to a layer?  End-users might expect to find that layer when they run `crane` or some similar tool.

```suggestion
A stack restricts buildpacks to run on top of a set of related **build images** (where your app is built)
and **run images** (where your app runs).  Practical experience of the buildpacks community has not found that the stack concept adds any value therefore [the community decided](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md)
to use buildpacks with build and run images without the additional concept formerly known as _stack_. 
```

content/docs/concepts/components/stack.md
```diff
@@ -6,7 +6,48 @@ aliases=[
...
+Stacks will continue to be supported with full backwards compatibility.
+
+### What's Next?
+[Targets](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md#example-buildpacktoml-targets-table) replace stacks.
+Targets allow buildpack authors to directly specify details such as OS and architecture directly without the intermediate object known as a "stack." 
+Targets are available starting with the 0.12 Platform API but should be safely ignored by older platforms.
+
+
+#### Wait - are you cutting me off?
+My friend, never fear. We have a long horizon and a heartfelt desire to ease your transition. We will provide backwards compatibility with stacks for so long that we haven't even decided yet when we're going to turn them off.
```
AidanDelaney, line 26: I like the more personal style that you use.  However we're much more straight-laced in the rest of the docs.  Should we prefer the (more boring) existing style?

```suggestion
My friend, never fear. We have a long horizon and a heartfelt desire to ease your transition. We will provide backwards compatibility with stacks for so long that we haven't even decided yet when we're going to turn them off.
```

AidanDelaney (changes requested):
The content here is great.  I've suggested some stylistic changes.  Feel free to adopt or ignore them.

AidanDelaney:
In answer to your question about https://buildpacks.io/docs/operator-guide/create-a-stack/, we should remove that.  The docs should reflect the current best practice.  I wonder if we should trust archive.org to store our legacy docs?  Or should we have a Buildpacks API dropdown at the top?

content/docs/concepts/components/stack.md
joe-kimmel-vmw: oh yes, almost certainly.

joe-kimmel-vmw:
> I wonder if we should trust archive.org to store our legacy docs?

Probably safe -- [it's not like a group of publishers would ever try to shut down the entire internet archive](https://www.npr.org/2023/03/26/1166101459/internet-archive-lawsuit-books-library-publishers).

content/docs/concepts/components/stack.md
```diff
@@ -6,7 +6,47 @@ aliases=[
...
+
+#### What is the current recommended best practice?
+We currently recommend that buildpack authors use both a stack and a target.
+We anticipate it will take some time before all platforms catch up with this change, but by using both your buildpacks will work with all platforms past, present, and future.
+
+In order to ease this process for those using the io.buildpacks.stacks.bionic, lifecycle will translate any section that sets this as on of the stacks:
+
+```
+[[stacks]]
+id = "io.buildpacks.stacks.bionic
```
jabrown85, line 36:
```suggestion
id = "io.buildpacks.stacks.bionic"
```

content/docs/concepts/components/target.md
```diff
@@ -0,0 +1,26 @@
...
+weight=4
+aliases=[
+    "/docs/using-pack/targets/"
+]
++++
+
+## What is a Target?
+
+A target is the essential information about an operating system and architecture necessary to ensure that binaries will be placed in environments where they can execute successfully.
+A buildpack may specify one or several targets. A build image or run image must specific one target.
```
jericop, line 12: This statement is confusing to me: `A build image or run image must specific one target.`. Does this mean `[[targets]]` is supposed to be nested inside the `[run]` and `[build]` blocks? I'm more familiar with yaml/json than with toml, so I tend to convert it in order to understand what is going on and it appears to be a builder-wide setting.

given this `builder.toml`:
```toml
[build]
image = "ghcr.io/jericop/build-jammy:target"

[run]
[[run.images]]
image = "ghcr.io/jericop/run-jammy:target"

[[targets]]
os = "linux"
arch = "amd64"
[[targets.distributions]]
name = "ubuntu"
versions = ["18.04", "20.04"]
```

the yaml output is:

```yaml
build:
  image: ghcr.io/jericop/build-jammy:target
run:
  images:
    - image: ghcr.io/jericop/run-jammy:target
targets:
  - os: linux
    arch: amd64
    distributions:
      - name: ubuntu
        versions:
          - "18.04"
          - "20.04"
```

It would be helpful to have a complete example of using `[[targets]]` in a builder toml.

content/docs/concepts/components/stack.md
```diff
@@ -6,7 +6,47 @@ aliases=[
...
+### Deprecation
+A stack restricts buildpacks to run on top of a set of related **build images** (where your app is built)
+and **run images** (where your app runs).  Practical experience of the buildpacks community has not found that the stack concept adds any value therefore [the community decided](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md)
+to use buildpacks with build and run images without the additional concept formerly known as _stack_. 
+
+Stacks will continue to be supported with full backwards compatibility.
+
+### What's Next?
+[Targets](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md#example-buildpacktoml-targets-table) replace stacks.
+Targets allow buildpack authors to directly specify details such as OS and architecture directly without the intermediate object known as a "stack."
```
AidanDelaney, line 20:
```suggestion
Targets allow buildpack authors to directly specify details such as OS and architecture directly without the intermediate object known as a "stack".
```

content/docs/operator-guide/create-targets.md
```diff
@@ -13,55 +13,66 @@ Before we get started, make sure you've got the following installed:
 
 {{< download-button href="https://store.docker.com/search?type=edition&offering=community" color="blue" >}} Install Docker {{</>}}
 
+## Creating custom images
 
-## Creating a custom stack
-
-In this tutorial we will create a sample stack based on `Ubuntu Bionic`. To create a custom stack, we need to create customized build and run images. Let's see how we can do so!
-
+In this tutorial we will create sample build and run images based on `Ubuntu Jammy.`
```
AidanDelaney, line 18:
```suggestion
In this tutorial we will create sample build and run images based on `Ubuntu Jammy`.
```

content/docs/operator-guide/create-targets.md
```diff
@@ -13,55 +13,66 @@ Before we get started, make sure you've got the following installed:
...
+Similarly 
+
+**Labels (optional)**
+
+##### TODO / question:   I think the spec says that "the platform"  (as opposed to the build image author) should set the labels?
+https://github.com/buildpacks/spec/pull/335/files#diff-e603760990971da3f77be4bb8d77c3405098f006814fd8c054d2d15f395b8330R199
+should we mention them in this tutorial? 
+##### end TODO /question
+
+
```
AidanDelaney, line 53: What direction is this taking?  Can I help here?

AidanDelaney (changes requested):
Getting closer :)

content/docs/operator-guide/create-targets.md
joe-kimmel-vmw: hmm. this is wrong as -is.   the "image config" doesn't get the `*` but the labels _do_, i think.   maybe I can ask @natalieparellano to confirm

joe-kimmel-vmw: ah, so, we said,  heck with it let's just not mention it in the tutorial since it's optional anyways, and only getting more optional as the spec gets closer and closer to being released.

natalieparellano:
Superseded by https://github.com/buildpacks/docs/pull/618

PR #591, opened 2023-06-10
fix: add missing .group to inline script

josegonzalez:
Without this change, the inline buildpack fails to be executed.


AidanDelaney (approved):
Thank you very much for this!

PR #598, opened 2023-07-12
Gather Build an App platform guides

AidanDelaney:
We have a different guide of Linux/amd64, Linux/arm64 and Windows/amd64.  Gather these together on a single tabbed page but default to the Linux/amd64 guide.


AidanDelaney:
#658 does this better.

PR #601, opened 2023-07-20
Update reference to Bionic in Stack tutorial

tcdowney:
Small change to make this tutorial mention `Jammy` instead of `Bionic` since that is what the rest of the doc references.


AidanDelaney:
@tcdowney Thank you for this.  Apologies for not merging it sooner, but we needed to fix an issue in CI first.

AidanDelaney (approved):
Finally! We had some CI issues. Thanks for this change.

PR #603, opened 2023-07-28
Use latest pack release 0.29.0

AidanDelaney:
Bump to the latest pack which supports extensions


AidanDelaney:
The issue here is lifecycle RC access.  Will address in a different PR.

PR #581, opened 2023-04-04
Demo run image extension

natalieparellano:

PR #604, opened 2023-07-28
Fix extensions demo now that samples repo has been updated

natalieparellano:
We don't need to re-create the builder, we just need to update the env vars passed to the buildpack to make it require things from extensions.

I wanted to make this fix as part of https://github.com/buildpacks/docs/pull/581, but that PR is blocked on there being new versions of pack and the lifecycle.


natalieparellano:
This also appears to be blocked on upgrading the lifecycle to at least 0.17.0-xxx due to the inclusion of `cowsay` extension in the samples repo

AidanDelaney (approved):
LGTM!

PR #605, opened 2023-08-03
Use pre-release version of pack for docs

AidanDelaney:
Use the latest pre-release of pack so that we can build images that use extensions API


AidanDelaney:
No longer necessary

PR #618, opened 2023-09-21
Add targets and deprecate stacks

natalieparellano:


content/docs/extension-guide/create-extension/setup-local-environment.md
```diff
@@ -72,7 +72,10 @@ We unset any custom lifecycle image that may have been provided to ensure that t
 
 <!-- test:exec -->
 ```bash
-git clone https://github.com/buildpacks/samples.git
+git clone https://github.com/buildpacks/samples.git && \
+  cd samples && \
+  git fetch --all && \
+  git checkout fix/curl
```
natalieparellano, line 78:
```suggestion
git clone https://github.com/buildpacks/samples.git
```
When https://github.com/buildpacks/samples/pull/176 is merged we can remove these

content/docs/reference/spec/buildpack-api.md
```diff
@@ -184,9 +189,31 @@ The schema is as follows:
...
 
-- **`stacks`** _(list, optional)_\
+- **`targets`** _(list, optional)_\
+  A list of targets supported by the buildpack.
+  _Cannot be used in conjunction with `order` list._
+
+  - **`os`** _(string, optional)_\
+    The supported operating system name.
+
+  - **`arch`** _(string, optional)_\
```
jjbustamante, line 199: I am a little confused, **when** `targets` are used, all the attributes are optional? `os` must be mandatory right?

content/docs/reference/spec/buildpack-api.md
```diff
@@ -184,9 +189,31 @@ The schema is as follows:
         - **`uri`** _(string, optional)_\
         A URL or path to the license.
 
-- **`stacks`** _(list, optional)_\
+- **`targets`** _(list, optional)_\
+  A list of targets supported by the buildpack.
```
natalieparellano, line 193:
```suggestion
  A list of targets supported by the buildpack.
  When no targets are specified, the `os`/`arch` will be inferred from the contents of the `./bin` directory
  (`./bin/build` implies `linux`/`amd64` and `./bin/build.bat` implies `windows`/`amd64`).
  For each target, all fields are optional (though at least one should be provided).
```

content/docs/reference/spec/buildpack-api.md
natalieparellano: In the end we decided that even `os` could be optional - though practically speaking, it will probably always be specified. I added some clarifying language here: https://github.com/buildpacks/docs/pull/618/files#r1334413773

natalieparellano:
@AidanDelaney this one should be ready and will fix CI

AidanDelaney (approved):

AidanDelaney (approved):
LGTM

PR #633, opened 2023-11-27
docs: update current buildpack version

jose-fully-ported:
This should match the version in https://github.com/buildpacks/spec/blob/buildpack/v0.10/buildpack.md


AidanDelaney:
Thanks for this.  Those docs _should_ be auto-generated.  I'll take a look into why they're not in the morning.

AidanDelaney (approved):
LGTM

PR #646, opened 2023-12-29
Updates hello world response documentation to reflect JSON

gpstines:
This fixes a small docs typo I found while browsing.


AidanDelaney (approved):
Thanks a lot!  We really appreciate it when people care enough to fix these small issues.

PR #649, opened 2024-01-24
Patch 1 -Fixed Issue in building-blocks-cnb.md 

thiunuwan:
Update documentation: replace 'Ruby' with 'Node.js'

Changed the word 'Ruby' to 'Node.js' in the documentation to correct an inaccuracy with the previous content.


AidanDelaney (approved):
Awesome! Thanks a lot.

PR #663, opened 2024-02-29
Move all images into static folder

natalieparellano:
Delete images that are duplicated in static folder and update their references

As a follow on to https://github.com/buildpacks/docs/pull/661 this will encourage all images to live in static and be referenced there.


AidanDelaney:
I'm not sure if we _need_ to do this.  I did it with the previous PR as only some of the svgs were being moved to `public` on Linux (whereas all were on macos).

natalieparellano:
Fair enough... but do we know __why__ it is needed sometimes and not others? My aim here is to encourage a pattern that is always going to work.

AidanDelaney (approved):
LGTM

PR #664, opened 2024-02-29
Consolidate common concepts

natalieparellano:
Uses the hugo union filesystem (what is recommended) instead of symlinks


AidanDelaney (approved):
LGTM

PR #666, opened 2024-03-06
Add app dev how-to for specifying export target along with diagram

natalieparellano:
showing how pack uses the daemon.


AidanDelaney (approved):
I like the diagram

PR #680, opened 2024-03-28
Update project description on front page

AidanDelaney:
We have developed a succinct description of what the project does. Include this on the front page before our claims about our heritage.

### DIFF
--- a/themes/buildpacks/layouts/index.html
+++ b/themes/buildpacks/layouts/index.html
@@ -90,7 +90,18 @@
     <a href="/features">See list of features</a>
   </div>
 
-  <h2>About the Project</h2>
+  <h2>What We Do</h2>
+  <section class="row">
+    <div class="col-sm">
+      <p>Cloud Native Buildpacks (CNBs) transform your application source code into container images that can run on any
+        cloud. With buildpacks, organizations can concentrate the knowledge of container build best practices within a
+        specialized team, instead of having application developers across the organization individually maintain their
+        own Dockerfiles. This makes it easier to know what is inside application images, enforce security and compliance
+        requirements, and perform upgrades with minimal effort and intervention.</p>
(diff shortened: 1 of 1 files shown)


samj1912 (approved):

themes/buildpacks/layouts/index.html
```diff
@@ -90,7 +90,18 @@ <h3 class="mb-0">Maintainability</h3>
...
   </div>
 
-  <h2>About the Project</h2>
+  <h2>What We Do</h2>
+  <section class="row">
+    <div class="col-sm">
+      <p>Cloud Native Buildpacks (CNBs) transform your application source code into container images that can run on any
+        cloud. With buildpacks, organizations can concentrate the knowledge of container build best practices within a
+        specialized team, instead of having application developers across the organization individually maintain their
+        own Dockerfiles. This makes it easier to know what is inside application images, enforce security and compliance
```
edmorley, line 99: Thank you for your work on improving the docs :-)

Reading the new description, it covers the "companies writing their own CNBs internally" case quite well, but I wonder if it's missing a mention of how for many people CNBs will actually be more about using third party tooling and not having to write their own at all? (eg those using the Paketo or Heroku language CNBs instead of writing their own Dockerfile)

natalieparellano: Maybe we could add an additional paragraph like, "The CNB project maintains a set of specifications and tooling that make it possible to write buildpacks and orchestrate them together in a build system. Additionally, community providers such as Google, Heroku, and the Paketo project maintain buildpacks for a wide variety of language families, and you can discover even more by searching the buildpack registry." WDYT?

edmorley: @natalieparellano Yeah something like that sounds great :-)

PR #682, opened 2024-03-28
Fix reference to OCI layout feature in pack

natalieparellano:
There isn't a --layout flag, you invoke layout with the image name.


AidanDelaney (approved):
LGTM

PR #684, opened 2024-04-04
Fix summaries for "custom env" page and add link to podman from main …

natalieparellano:
…tutorial

I re-worded the main tutorial to indicate that podman is an alternative to docker

Also in looking into this, I noticed that the summary page for "build in custom environments" looked a little off, see before and after:

Screenshot 2024-04-04 at 11 04 10 AM

Screenshot 2024-04-04 at 11 04 44 AM


AidanDelaney (approved):
LGTM

Issue #694, opened 2024-04-19
Bug: Edit Page Redirects to Incorrect File Path


mhmohona:
## Summary
When attempting to edit the documentation pages for both buildpacks and composite buildpacks at https://buildpacks.io/docs/for-app-developers/concepts/buildpack/ and https://buildpacks.io/docs/for-platform-operators/concepts/composite-buildpack/, the links redirect to non-existent file paths, resulting in a [`File not found`](https://github.com/buildpacks/docs/edit/main/content/docs/for-app-developers/concepts/buildpack.md?description=Signed-off-by%3A+NAME+%3CEMAIL_ADDRESS%3E%0A%0A%3E+_By+signing+off+you+acknowledge+adhering+to+the+requirements+listed+here%3A+https%3A%2F%2Fprobot.github.io%2Fapps%2Fdco%2F_) error.

## What questions should this documentation answer?
- 

## Who is the target audience for this documentation?

- [x] App Developer
- [x] Buildpack Author
- [x] Operator
- [ ] Other:

**Steps to Reproduce:**
1. Navigate to the documentation page for buildpacks at https://buildpacks.io/docs/for-app-developers/concepts/buildpack/.
2. Click on the "Edit this page" link.
3. Observe the redirection to https://github.com/buildpacks/docs/edit/main/content/docs/for-app-developers/concepts/buildpack.md, which leads to a "File not found" error.
4. Navigate to the documentation page for composite buildpacks at https://buildpacks.io/docs/for-platform-operators/concepts/composite-buildpack/.
5. Click on the "Edit this page" link.
6. Observe the redirection to a non-existent file path, which leads to a "File not found" error.

**Expected Behavior:**
The "Edit this page" link should redirect to the correct file path where the documentation for [buildpacks](https://github.com/buildpacks/docs/blob/main/content/docs/.common/concepts/buildpack.md) and [composite buildpacks](https://github.com/buildpacks/docs/blob/main/content/docs/.common/concepts/composite-buildpack.md) is located.

**Actual Behavior:**
The links redirect to non-existent file paths, causing a "File not found" error.

PR #696, opened 2024-04-20
Fix edit links for buildpacks documentation

MeenuyD:
The edit links for the documentation pages were redirecting to non-existent file paths, resulting in a File not found error. This commit corrects the edit links to point to the correct file paths, allowing contributors to edit the documentation pages for buildpacks directly from the website.
#694


AidanDelaney (changes requested):
Could you please explain how this fixes the linking problem?

AidanDelaney:
Closing until we can explain the problem that is fixed by the PR

PR #702, opened 2024-05-01
[WIP] Use fully qualified image names

hhiroshell:
Addresses the https://github.com/buildpacks/pack/issues/1218, specifically the modifications to the documentation.

This PR changes the container image names used in the document to fully qualified names. Currently, I'm adopting the following policies:

- Unless there is a specific reason not to, use `registry.example.com` for the hostname and `example` for the organization name. 
  - Example: `registry.example.com/example/my-app`
- In tutorial pages or similar contexts where commands are expected to work actually, use an implied hostname.
  - Example: use `docker.io/ubuntu:jammy` instead of `ubuntu:jammy`
- Do not add a remote registry name to an image that is built and used only within a local environment.

### DIFF
--- a/content/docs/for-app-developers/concepts/base-images/stack.md
+++ b/content/docs/for-app-developers/concepts/base-images/stack.md
@@ -42,8 +42,8 @@ Stacks are used by [builders][builder] and are configured through a builder's
 
 [stack]
   id = "com.example.stack"
-  build-image = "example/build"
-  run-image = "example/run"
+  build-image = "docker.io/example/build"
+  run-image = "docker.io/example/run"
   run-image-mirrors = ["gcr.io/example/run", "registry.example.com/example/run"]
 ```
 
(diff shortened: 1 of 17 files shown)


content/docs/for-app-developers/concepts/base-images/stack.md
```diff
@@ -42,8 +42,8 @@ Stacks are used by [builders][builder] and are configured through a builder's
 
 [stack]
   id = "com.example.stack"
-  build-image = "example/build"
-  run-image = "example/run"
+  build-image = "docker.io/example/build"
```
AidanDelaney, line 45: Does this image need a registry?  I think this may meet your criteria of only being used in the local environment

hhiroshell: In my understanding, this image means it is pulled from a remote registry to be used as a build environment, so I thought it should be specified with the registry name.

natalieparellano: This should be okay, the daemon will happily accept `some-image` and `index.docker.io/some-image` interchangeably.

Side note, while `docker.io/some-image` is correct, all of our code will resolve a registry-less image name to `index.docker.io/<image>` so I think using the fully qualified domain would be clearer.

PR #703, opened 2024-05-03
update to run the `pack` command

laurentheirendt:
The previous command is not runnable as the location of the command is important.

Alternative when outside the `node-js-sample-app` directory:
```bash
pack build test-node-js-app --path ./node-js-sample-app --buildpack ./node-js-sample-app/node-js-buildpack
```


content/docs/for-buildpack-authors/tutorials/basic-buildpack/02_building-blocks-cnb.md
```diff
@@ -113,11 +113,11 @@ pack config trusted-builders add cnbs/sample-builder:jammy
 ```
 <!--+- "{{execute}}"+-->
 
-Then run the following `pack` command:
+Then run the following `pack` command after `cd` to `node-js-sample-app`:
```
natalieparellano, line 116:
```suggestion
Then run the following `pack` command from the `node-js-sample-app` directory:
```

PR #706, opened 2024-05-04
change pack command

laurentheirendt:
similar to #703


natalieparellano:
Looking at the failing test, I'm a bit confused as to why the original command didn't work. I think we expect the buildpack and the app directories to be siblings, not one the child of the other. Is that not clear (or incorrect) in the tutorial instructions?

laurentheirendt:
The command doesn't work because the command needs to be either run from within the directory `node-js-sample-app` or from outside of it. Here is the directory structure:

```
.
└── node-js-sample-app
    ├── app.js
    ├── node-js-buildpack
    │   ├── bin
    │   │   ├── build
    │   │   └── detect
    │   └── buildpack.toml
    └── package.json

4 directories, 5 files
```

natalieparellano:
Ah, you know, looking again at the buildpack author tutorial, I think we expect the directory structure to be like this:

```
└── node-js-sample-app
    ├── app.js
    └── package.json
└── node-js-buildpack
    ├── bin
    │   ├── build
    │   ├── detect
    └── buildpack.toml
```

So this page https://buildpacks.io/docs/for-buildpack-authors/tutorials/basic-buildpack/01_setup-local-environment/ doesn't actually `cd` into the app directory, but it does tell you to create a file there. If you look at the test code that gets executed ([here](https://github.com/buildpacks/docs/blob/main/content/docs/for-buildpack-authors/tutorials/basic-buildpack/01_setup-local-environment.md?plain=1#L40-L42)) it is in the context of the top-level directory.

I think we can certainly make the tutorial clearer, so that others don't run into the same issue. Maybe we could make the line that says `Create a file in the current directory called node-js-sample-app/app.js with the following contents:` into a copy-pastable command that doesn't change the directory?

laurentheirendt:
Ah, that makes sense then 😄 

I think your suggestion would also work 👍🏼

natalieparellano (approved):

AidanDelaney (approved):
LGTM

PR #715, opened 2024-05-23
Update arm guide with arm-compatible builder instructions

runesoerensen:
There's finally a multi-arch builder publicly available, so this guide can be simplified substantially. It might also make sense to merge this with some of the content in this guide https://buildpacks.io/docs/for-app-developers/tutorials/basic-app/ at some point.


AidanDelaney (approved):
This is awesome! I think our CI is currently broken..so it'll take me a day or so to land this.

PR #717, opened 2024-05-29
Fix references to "meta buildpack"

natalieparellano:
It should be called a "composite buildpack" but we can refer to the old name too


AidanDelaney (approved):
LGTM

PR #723, opened 2024-06-05
Update ARM build docs

runesoerensen:
Updating the ARM build docs to reflect [`pack build --platform`](https://github.com/buildpacks/pack/pull/2162) support introduced in [`pack v0.34.0`](https://github.com/buildpacks/pack/releases/tag/v0.34.0).

### DIFF
--- a/content/docs/for-app-developers/how-to/special-cases/build-for-arm.md
+++ b/content/docs/for-app-developers/how-to/special-cases/build-for-arm.md
@@ -10,7 +10,7 @@ weight=1
 <!--more-->
 
 Building for the ARM architecture is now easier than ever! The `heroku/builder:24` builder supports both AMD64 and ARM64 architectures, and includes
-multi-arch Go, Java, Node.js, Python, Ruby and Scala buildpacks. You can read more about Heroku's [Cloud Native Buildpacks here][heroku-buildpacks].
+multi-arch Go, Java, Node.js, PHP, Python, Ruby and Scala buildpacks. You can read more about Heroku's [Cloud Native Buildpacks here][heroku-buildpacks].
 
 ### 1. Clone the [samples][samples] repository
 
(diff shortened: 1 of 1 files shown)


content/docs/for-app-developers/how-to/special-cases/build-for-arm.md
```diff
@@ -22,17 +22,17 @@ git clone https://github.com/buildpacks/samples
...
 ```
 pack build java-maven-sample --path samples/apps/java-maven/ --builder heroku/builder:24
 ```
 <!--+- "{{execute}}"+-->
 
-As `heroku/builder:24` is a multi-arch builder, it'll default to the current architecture, and an AMD64 image will be built when running `pack` on that architecture.
+By default, `pack` uses the current architecture for multi-arch builders like `heroku/builder:24`, so an AMD64 image will be built on AMD64 systems.
 
-If you want to build an ARM64 image from a different host architecture, you can use the arch-specific builder tag: `heroku/builder:24_linux-arm64`:
+If you want to build an ARM64 image from a different host architecture, use the `--platform` parameter:
```
AidanDelaney, line 33: Does building a `linux/arm64` image on a `linux/amd64` host would require some form of `arm64` emulator?  I know we can assemble the images.  But does building work in the general case?  For example, if I'm building a Python application that, in turn, needs to compile a C library?

runesoerensen: If I understand you correctly, I think the answer to both your questions is yes -- building an `linux/arm64` image on a `linux/amd64` host does require an emulator capable of emulating/virtualizing a different architecture (like Docker :)), and building should generally work - also if a user needs to compile a C library.

Running `pack build` with the `--builder heroku/builder:24 --platform linux/arm64` parameters is equivalent to using the arch-specific `heroku/builder:24_linux-arm64` builder tag currently in the docs, and the builder will be the same either way.

However, the only reason to use `--builder heroku/builder:24_linux-arm64` (and the reason we added arch-specific tags) was that `pack build` didn't allow specifying a `--platform` for multi-arch builders until just recently. It was a bit of a hack which, since the introduction of pack v0.34.0, is fortunately no longer necessary! :)

AidanDelaney: Ah great.  I'll land this.  But it would be also nice to have a list of builders that support ARM in this doc (simply because ARM support is still a little rare).

AidanDelaney (approved):
LGTM

PR #727, opened 2024-06-13
fix build-base example

joeybrown-sf:
These extra slashes result in lines being unintentionally appended to previous `RUN` statements. This breaks the image build.

Removing them gives us the desired behavior.


AidanDelaney (approved):
LGTM

PR #731, opened 2024-06-18
docs: note python3-launchpadlib requirement for Debian (Ubuntu) pack-cli install

whalderman:
`sudo add-apt-repository ppa:cncf-buildpacks/pack-cli` fails with an unhelpful error if `python3-launchpadlib` is not already installed on the system.

> AttributeError: 'NoneType' object has no attribute 'people'


whalderman:
I got ahead of myself, the repository doesn't support Debian.

PR #761, opened 2024-08-22
Document builder property available in project descriptor

mars:
I found some [Google Cloud docs about this](https://cloud.google.com/docs/buildpacks/use-a-specific-builder), but they seem to be non-standard. This change adds a section explaining the standard way to set the builder.

### DIFF
--- a/content/docs/for-app-developers/how-to/build-inputs/use-project-toml.md
+++ b/content/docs/for-app-developers/how-to/build-inputs/use-project-toml.md
@@ -107,6 +107,21 @@ pack build sample-app \
 docker run sample-app
 ```
 
+### Specify Builder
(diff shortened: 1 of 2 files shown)


edmorley (commented):
Thank you for this!

I don't suppose you'd mind also documenting the builder option on:
https://buildpacks.io/docs/reference/config/project-descriptor/

(It seems https://buildpacks.io/docs/reference/config/project-descriptor/ is missing a few things from https://github.com/buildpacks/spec/blob/main/extensions/project-descriptor.md)

mars:
Done @edmorley 🙌

AidanDelaney (approved):
This is great! Thank you for your contribution.

PR #762, opened 2024-08-27
Update specify-launch-process docs

runesoerensen:
Remove mention of `pack` always treating `web` as the default process type. The `java-maven` buildpack [was updated](https://github.com/buildpacks/samples/commit/8fa4447e65145e5a727d203103c554bce2b085d6) to explicitly set the default process type (after the buildpack API version was bumped from 0.2 to 0.9, and [`pack` only sets the `web` process as default for < 0.6](https://github.com/buildpacks/spec/blob/main/platform.md#outputs-4)).

Also, the java-maven sample buildpack version is now [0.0.2](https://github.com/buildpacks/samples/commit/09460d909bc9f793e173e384f7396d4adbb07f3c).

### DIFF
--- a/content/docs/for-app-developers/how-to/build-outputs/specify-launch-process.md
+++ b/content/docs/for-app-developers/how-to/build-outputs/specify-launch-process.md
@@ -54,7 +54,7 @@ Rebasable: true
 
 Buildpacks:
   ID                             VERSION           HOMEPAGE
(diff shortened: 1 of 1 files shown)


AidanDelaney (approved):
Thanks for this. I really  appreciate your attention to detail!

PR #770, opened 2024-09-28
Adds Build time and runtime images content to the Platform docs

hyounes4560:
- adds build and run images content under the platform section

### DIFF
--- a/content/docs/for-platform-operators/concepts/build-run-images.md
+++ b/content/docs/for-platform-operators/concepts/build-run-images.md
@@ -1,0 +1,94 @@
++++
+title="Base image types"
+weight=99
++++
+
+As you already know, `Cloud Native Buildpacks (CNBs)` transform your application source code into `OCI images`  that can run on any cloud and include all of the components your application needs to build and run and.
+
+<!--more-->
+
+Each buildpack checks the source code and provides any relevant dependencies. Then, a `base image` is then generated from the application’s source code and these dependencies.
+
+## Base image types
+
+A `base image` is an `OCI image` containing the base, or initial set of layers, for other images. It is helpful to distinguish between two distinct types of images, `Build` and `Run` images.
+
+### Build image
+
+A `build image` is an `OCI image` that serves as the base image for the `build` environment in which buildpacks `lifecycle` is executed.
+
+A typical `build image` might determine:
+
+* The OS distro in the build environment
+* OS packages installed in the build environment
+* Trusted CA certificates in the build environment
+* The default user in the build environment
+
+The platform must ensure that:
+
+* The image config's `User` field is set to a non-root user with a writable home directory
+* The image config's `Env` field has the environment variable `CNB_USER_ID` set to the user [UID/SID](https://github.com/buildpacks/spec/blob/main/README.md#operating-system-conventions) of the user specified in the `User` field
+* The image config's `Env` field has the environment variable `CNB_GROUP_ID` set to the primary group [GID/SID](https://github.com/buildpacks/spec/blob/main/README.md#operating-system-conventions) of the user specified in the `User` field
+* The image config's `Env` field has the environment variable `PATH` set to a valid set of paths or explicitly set to empty (`PATH=`)
+
+The platform should ensure that:
+
+* The image config's `Label` field has the label `io.buildpacks.base.maintainer` set to the name of the image maintainer.
+* The image config's `Label` field has the label `io.buildpacks.base.homepage` set to the homepage of the image.
+* The image config's `Label` field has the label `io.buildpacks.base.released` set to the release date of the image.
+* The image config's `Label` field has the label `io.buildpacks.base.description` set to the description of the image.
+* The image config's `Label` field has the label `io.buildpacks.base.metadata` set to additional metadata related to the image.
+
+#### Anatomy of a build image
+
+Typically, a `build` image may include:
+
+* Shell
+* C-compiler
+* Minimal operating system distribution, such as Linux utilities that build systems might call out to
+* Build time libraries
+
+### Run image
+
+A `run image` is an `OCI image` that serves as the base image for the final application image.
+
+A typical run image might determine:
+
+* The OS distro or distroless OS in the launch environment
+* OS packages installed in the launch environment
+* Trusted CA certificates in the launch environment
+* The default user in the run environment
+
+The platform must ensure that:
+
+* The image config's `Env` field has the environment variable `PATH` set to a valid set of paths or explicitly set to empty (`PATH=`)
+
+The platform should ensure that:
+
+* The image config's `User` field is set to a user with a **DIFFERENT** user [UID/SID](https://github.com/buildpacks/spec/blob/main/README.md#operating-system-conventions) as the build image
+* The image config's `Label` field has the label `io.buildpacks.base.maintainer` set to the name of the image maintainer
+* The image config's `Label` field has the label `io.buildpacks.base.homepage` set to the homepage of the image
+* The image config's `Label` field has the label `io.buildpacks.base.released` set to the release date of the image.
+* The image config's `Label` field has the label `io.buildpacks.base.description` set to the description of the image
+* The image config's `Label` field has the label `io.buildpacks.base.metadata` set to additional metadata related to the image
+* The image config's `Label` field has the label `io.buildpacks.rebasable` set to `true` to indicate that new run image versions maintain [ABI-compatibility](https://en.wikipedia.org/wiki/Application_binary_interface) with previous versions (see [Compatibility Guarantees]([#compatibility-guarantees](https://github.com/buildpacks/spec/blob/main/platform.md#compatibility-guarantees))).
+
+#### Anatomy of a run image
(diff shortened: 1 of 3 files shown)


content/docs/for-platform-operators/concepts/build-run-images.md
```diff
@@ -0,0 +1,94 @@
++++
+title="Base image types"
+weight=99
++++
+
+As you already know, `Cloud Native Buildpacks (CNBs)` transform your application source code into `OCI images`  that can run on any cloud and include all of the components your application needs to build and run and.
+
+<!--more-->
+
+Each buildpack checks the source code and provides any relevant dependencies. Then, a `base image` is then generated from the application’s source code and these dependencies.
```
AidanDelaney, line 10: Is "base image" the correct term here?  We've used "application image" in the past to refer to the final output image.

content/docs/for-platform-operators/concepts/build-run-images.md
```diff
@@ -0,0 +1,94 @@
...
+
+* The OS distro in the build environment
+* OS packages installed in the build environment
+* Trusted CA certificates in the build environment
+* The default user in the build environment
+
+The platform must ensure that:
+
+* The image config's `User` field is set to a non-root user with a writable home directory
+* The image config's `Env` field has the environment variable `CNB_USER_ID` set to the user [UID/SID](https://github.com/buildpacks/spec/blob/main/README.md#operating-system-conventions) of the user specified in the `User` field
```
AidanDelaney, line 30: We may have to explain what is meant by "`User` field" and "`Env` field".  Do we mean user and environment here?  i.e. that the build image must have a non-root user and `CNB_USER_ID` is an environment variable which contains the `uid` of that user.

content/docs/for-platform-operators/concepts/build-run-images.md
```diff
@@ -0,0 +1,94 @@
...
+A typical run image might determine:
+
+* The OS distro or distroless OS in the launch environment
+* OS packages installed in the launch environment
+* Trusted CA certificates in the launch environment
+* The default user in the run environment
+
+The platform must ensure that:
+
+* The image config's `Env` field has the environment variable `PATH` set to a valid set of paths or explicitly set to empty (`PATH=`)
```
AidanDelaney, line 64: Again, I wonder if "`Env` field" is the correct term here.  Do we mean that the run image must define a `PATH` environment variable?

content/docs/for-platform-operators/concepts/build-run-images.md
```diff
@@ -0,0 +1,94 @@
...
+* The image config's `Label` field has the label `io.buildpacks.base.maintainer` set to the name of the image maintainer
+* The image config's `Label` field has the label `io.buildpacks.base.homepage` set to the homepage of the image
+* The image config's `Label` field has the label `io.buildpacks.base.released` set to the release date of the image.
+* The image config's `Label` field has the label `io.buildpacks.base.description` set to the description of the image
+* The image config's `Label` field has the label `io.buildpacks.base.metadata` set to additional metadata related to the image
+* The image config's `Label` field has the label `io.buildpacks.rebasable` set to `true` to indicate that new run image versions maintain [ABI-compatibility](https://en.wikipedia.org/wiki/Application_binary_interface) with previous versions (see [Compatibility Guarantees](https://github.com/buildpacks/spec/blob/main/platform.md#compatibility-guarantees)).
+
+#### Anatomy of a run image
+
+A `runtime` image may contain:
```
AidanDelaney, line 78: What's the difference between the "run image" and a "runtime" image?

AidanDelaney (changes requested):
Mostly questions about terminology.

content/docs/for-platform-operators/concepts/build-run-images.md
natalieparellano, line 10:
```suggestion
Each buildpack checks the source code and provides any relevant dependencies in the form of layers. Then, buildpack-provided layers are placed atop a runtime `base image` to form the final application image.
```

content/docs/for-platform-operators/concepts/build-run-images.md
```diff
@@ -0,0 +1,94 @@
...
+
+Each buildpack checks the source code and provides any relevant dependencies. Then, a `base image` is then generated from the application’s source code and these dependencies.
+
+## Base image types
+
+A `base image` is an `OCI image` containing the base, or initial set of layers, for other images. It is helpful to distinguish between two distinct types of images, `Build` and `Run` images.
+
+### Build image
+
+A `build image` is an `OCI image` that serves as the base image for the `build` environment in which buildpacks `lifecycle` is executed.
```
natalieparellano, line 18:
```suggestion
A `build image` is an `OCI image` that serves as the base image for the `build` environment in which the CNB `lifecycle` and buildpacks are executed.
```

content/docs/for-platform-operators/concepts/build-run-images.md
natalieparellano: I think nothing... I think we decided to use "runtime base image" as opposed to "run image" because the latter was more of a CNB construct whereas the former would be recognized more widely.

content/docs/for-platform-operators/concepts/build-run-images.md
natalieparellano: In general, I think we should avoid copy/pasting the spec into the docs. Then we just have two things to maintain. This section should provide some grounding for understanding the spec, and then link to it. Let's think about how to frame this and then provide a code suggestion.

content/docs/for-platform-operators/concepts/build-run-images.md
```diff
@@ -0,0 +1,94 @@
++++
+title="Base image types"
```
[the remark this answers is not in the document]
hyounes4560: renamed

content/docs/for-platform-operators/concepts/build-run-images.md
hyounes4560: this is from the spec; not sure how valid it's

content/docs/for-platform-operators/concepts/build-run-images.md
hyounes4560: @natalieparellano then we would need something more broad that we could use. The spec is very specific and I can't change things as every line is very focused!! Do we have any other resources that I could use?

content/docs/for-platform-operators/concepts/build-run-images.md
AidanDelaney: The spec defines terminology at https://github.com/buildpacks/spec/blob/main/platform.md#terminology.

The terminology is defines *base image*, *build image*, *run image* and *app image*.  An *app image* a.k.a. *application image* is generated by combining the *run image* with the layers generated by executing a set of buildpacks against the application source code.

content/docs/for-platform-operators/concepts/build-run-images.md
AidanDelaney: We don't have other sources.  We really struggle to explain these things.  Let's chat about this in our regular catchup.  I think there's a straightforward solution here.  We may have to unpack the idea of an _image config_ using an example, so that readers get a clear idea of what these fields look like.

content/docs/for-platform-operators/concepts/build-run-images.md
hyounes4560: updated already

content/docs/for-platform-operators/concepts/build-run-images.md
hyounes4560: removed as per our latest sync

hyounes4560:
updated based on @natalieparellano feedback; please take a look. FYI @AidanDelaney

AidanDelaney (approved):
LGTM, waiting for https://github.com/buildpacks/docs/pull/773 to fix the linting problem.

natalieparellano (approved):

AidanDelaney:
Will fix the broken links (which are causing build filaure) later:

```
ERROR: found broken links:
target does not exist --- docs/for-platform-operators/concepts/builder/index.html --> /docs/for-platform-operators/concepts/base-images.md
```

content/docs/for-platform-operators/concepts/builder.md
```diff
@@ -20,14 +20,15 @@ A builder consists of the following components:
...
 ### Resources
 
 To learn how to create your own builder, see our [Operator's Guide][operator-guide].
 
-[builder-config]: /docs/reference/builder-config/
 [buildpack]: /docs/for-platform-operators/concepts/buildpack/
 [lifecycle]: /docs/for-platform-operators/concepts/lifecycle/
 [operator-guide]: /docs/for-platform-operators/
+[build-image]: /docs/for-platform-operators/concepts/base-images.md
+[run-image]: /docs/for-platform-operators/concepts/base-images.md
```
AidanDelaney, line 34:
```suggestion
[build-image]: /docs/for-platform-operators/concepts/base-images
[run-image]: /docs/for-platform-operators/concepts/base-images
```

PR #813, opened 2025-03-17
Add an explanation about performance degradation when using the Docker daemon with containerd storage

hhiroshell:
This PR adds an explanation about performance degradation when using the Docker daemon with containerd storage following the discussion in the https://github.com/buildpacks/pack/issues/2272 .
In the PR, we are moving forward in [4 steps](https://github.com/buildpacks/pack/issues/2272#issuecomment-2426435484). and this PR is for the step 2 .

Here is the change.

#### Target page
- https://buildpacks.io/docs/app-journey/

#### before
Image

#### after
Image


natalieparellano (approved):
Thank you @hhiroshell this is really helpful

content/docs/app-journey.md
```diff
@@ -95,6 +95,10 @@ deploying your new image to your favorite cloud!
 
 > In case you need it, `pack build` has a handy flag called `--publish` that will build your image directly onto a Docker
 > registry. You can learn more about `pack` features in the [documentation][pack-docs].
+>
+> Additionally, if the Docker daemon is using containerd storage and you export a container to the Docker daemon 
+> (building without `--publish`), it may degrade the performance of `pack build`. For more details, please refer to
+> [the related issue](https://github.com/buildpacks/pack/issues/2272).
```
AidanDelaney, line 101:
```suggestion
>
> Additionally, if your Docker daemon uses `containerd` image storage (not the Docker default),
> it may degrade the performance of `pack build`. For more details, please refer to
> [the related issue](https://github.com/buildpacks/pack/issues/2272).
```

AidanDelaney: Folks who configure `containerd` are likely to be advanced users.  I've modified the text so that casual users can understand that this problem exists only in non-default configurations.

hhiroshell: I think your suggestion is better than my original phrasing. Thanks!

natalieparellano: It's worth noting, containerd will become the default eventually. Maybe "not the Docker default as of version x.x.x"?

PR #821, opened 2025-04-07
Update arch install

bupd:
## Changes made
- Update install on Arch Linux.


natalieparellano (approved):
Looks good to me. Thanks for the PR @bupd

PR #841, opened 2025-07-28
Clear up `clean-env` docs

schneems:
## Context

I was asked to validate that several buildpacks were correctly passing user-supplied environment variables to sub-processes for the purpose of configuring access to dependencies that require credentials.

The docs clearly state that a `buildpack.toml` with `clear-env` should have specific behavior regarding environment variable visibility, but it's not clear that it means that the `CNB_PLATFORM_DIR/env/` is ALWAYS provided and that `clear-env = false` means that information is merged into the current environment.

If you read it literally and you know that `$CNB_PLATFORM_DIR/env/` is a mechanism to provide the environment variables to a buildpack, you might interpret it as the ability to hide these platform envs from the buildpack. When really, it's saying "lifecycle will make things easy for you if you just want a firehose of all the envs, it will copy the user-provided ones into the current environment so you don't need to." It does this by default with the ability to opt out by setting `clear-env = true`.

## Change

I'm suggesting we tighten up the language here to make it clear that `$CNB_PLATFORM_DIR/env/` is always available, and that setting `clear-env = true` will tell the lifecycle to copy the contents of `CNB_PLATFORM_DIR/env/` into the current environment. Specific language suggestions pending. This logic is in several places:

- https://buildpacks.io/docs/for-buildpack-authors/how-to/write-buildpacks/clear-env/
- https://buildpacks.io/docs/reference/config/buildpack-config/
- https://buildpacks.io/docs/for-app-developers/how-to/build-inputs/configure-build-time-environment/

### DIFF
--- a/content/docs/for-app-developers/how-to/build-inputs/configure-build-time-environment.md
+++ b/content/docs/for-app-developers/how-to/build-inputs/configure-build-time-environment.md
@@ -11,9 +11,9 @@ Environment variables are a common way to configure various buildpacks at build-
 
 <!--more-->
 
-Below are a few ways you can do so. All of them will use our [samples] repo for
-simplicity. The following documentation assumes that all participating buildpacks
-either use `clear-env = false` (the default) in their
+Below are a few ways you can do so. All of them use the [samples] repository for
+simplicity. The following documentation assumes that all participating buildpack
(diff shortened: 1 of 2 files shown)


AidanDelaney (approved):
This is a great improvement. Thank you.

PR #866, opened 2025-10-27
Fix formatting of kpack Donation announcement link

willkara:
May be pedantic, but the link wasn't being rendered correctly due to some annoying spacing issues.


AidanDelaney (approved):
LGTM

PR #870, opened 2025-11-20
Switch to pagefind for search

AidanDelaney:
We have more control over search indices if we use a local search library

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ build: $(HUGO_BIN) pack-version pack-docs-update
 	@echo "> Building..."
 	$(HUGO_BIN)
 	# Github Workflows ensures `node` is installed.
-	npx run pagefind --source ./public --destination ./public/pagefind
+	npx pagefind --site ./public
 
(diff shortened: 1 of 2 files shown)


AidanDelaney:
There are too many changes on this branch.  Will rebase once https://github.com/buildpacks/docs/pull/867 is landed

Makefile
```diff
@@ -30,17 +30,9 @@ TOOLS_BIN:=tools/bin
...
-HUGO_ARCH:=32bit
+HUGO_ARCH:=64bit
 HUGO_EXT:=tar.gz
-ifeq ($(OS),Windows_NT)
-	HUGO_OS:=Windows
-	HUGO_EXT:=zip
-ifeq ($(PROCESSOR_ARCHITECTURE),AMD64)
-	HUGO_ARCH:=64bit
-endif
-endif
```
jjbustamante: Looks good to me, my only question is: removing these lines means people in the community using Windows will not be able to compile and contribute to the docs repo?

